PR #239, opened 2021-06-21
add config points for user agent in header for loader requests

### LINKED ISSUE #93: Set a user-agent for all API requests
We should set a nice User-Agent header for all our requests to authorized APIs (CVS, Rite Aid, etc.).

Probably something like:

```
univaf-loader/<version> (+https://github.com/usdigitalresponse/appointment-availability-infra)
```

gsandstrom:
- ~set USER_AGENT string in .env~
- ~retain ability to use randomized user agent strings in CVS requests with the CVS_RANDOMIZE_USER_AGENT .env parameter~

~Still want to do a test before merging - seeing if I can capture the requests being made in the jest tests against the mock nock server.~

**EDIT: This got refactored significantly. (by @Mr0grog)**

This main change here is just that we now have `utils.httpClient`, which is a customized version of `got` with the correct `User-Agent` header. Anywhere we make HTTP requests, we *should* use `utils.httpClient` instead of `got` unless there’s a very special situation.

There is also a `RELEASE` environment variable that, in production, will have the commit hash the release was built from. If release isn’t set, it falls back to using the `version` field from `package.json`. This is used to fill in the version number in the User Agent string.

fixes #93

### DIFF
--- a/loader/test/utils.test.js
+++ b/loader/test/utils.test.js
@@ -35,8 +35,8 @@ describe("httpClient", () => {
         userAgent = this.req.headers["user-agent"];
         return [200, "OK"];
       });
-
     await httpClient({ url: "https://example.com/" });
+
     // The User Agent should be "univaf-loader/<version>" (where version is a
     // (git hash or "#.#.#"), optionally followed by a space and anything else.
     // Examples:


Mr0grog:
@gsandstrom is taking a break from this project, so I’ll work on finishing this PR.

Mr0grog:
First step: rebase and update files to pass prettier checks. Still need to update the code to go with the comments I made a while back, and to apply this to all the new loaders that have been written in the mean time.

PR #274, opened 2021-06-30
[233] Small Terraform Reorganization

jaronoff97:
This PR is some small clean up for terraform (mostly imports and references here and there.) It doesn't block on the public OSS stuff. Part of this PR was figuring out how to run terraform locally, this will allow us to upgrade to terraform 1.0.1 with some more ease, as well as do some refactoring of our modules to clean up.


terraform/remote.tf
```diff
@@ -0,0 +1,9 @@
+terraform {
+  backend "remote" {
+    organization = "usdr"
+
+    workspaces {
+      name = "appointment-availability-infra"
+    }
+  }
+}
```
Mr0grog, line 9: Now that this is here, what can or should we do to discourage running commands that might taint the production state? Do they always require manually locking the workspace, and is that clear enough?

(In reference to Slack discussion: https://usdigitalresponse.slack.com/archives/C01S4JWSMT7/p1625088086008200?thread_ts=1625024517.032900&cid=C01S4JWSMT7)

terraform/variables.tf
```diff
@@ -115,9 +110,14 @@ variable "loader_sentry_dsn" {
...
-  description = "ID for a security group in AWS that allows access to a bastion server inside our VPC. SSHing into this server will allow access to any services that allow this security group."
-  default     = "sg-06fc404762de692db"
+variable "bastion_key_name" {
+  description = "The key pair name to use for running the bastion server"
+  default     = "univaf-bastion-key"
+}
+
+variable "bastion_ami_group_id" {
+  description = "AMI to be used for the bastion server"
+  default = "ami-03d5c68bab01f3496"
```
Mr0grog, line 120: Are these in preparation for a bastion config that does not actually exist on this PR yet? I don’t see them actually referenced anywhere.

terraform/variables.tf
```diff
@@ -136,3 +136,11 @@ variable "data_snapshot_aws_secret_key" {
...
   sensitive = true
 }
+
+variable "AWS_SECRET_ACCESS_KEY" {
+  default = ""
+}
+
+variable "AWS_ACCESS_KEY_ID" {
+  default = ""
+}
```
Mr0grog, line 146: Necessary for local `terraform` to import/plan/apply, or is this just cleaning up the warning we currently get in all our plans in Terraform Cloud?

Mr0grog (commented):
Getting back around to this after all the mess from Monday night.

What’s the current plan for how the work here needs to continue? I *think* the changes to production state don’t have to happen until right before this merges, correct? If so, I think you should:
1. Hold off on those until you merge this.
2. Describe in this PR the Terraform commands that need to be run alongside/before merging.
3. Like we talked about on Slack, work out a plan and time window for doing the above and articulate it in this PR.

terraform/variables.tf
jaronoff97: Correct, this is just to clean up that warning which was preventing my local planning

terraform/variables.tf
jaronoff97: ah yeah this was when i was attempting to import the ec2 instance in to this, i should remove it.

astonm (approved):
Nice, all looks pretty good to me. One weird thing is that the plan seems to do something weird with the main security group:

```terraform
      ~ ingress                = [
          - {
              - cidr_blocks      = []
              - description      = ""
              - from_port        = 5432
              - ipv6_cidr_blocks = []
              - prefix_list_ids  = []
              - protocol         = "tcp"
              - security_groups  = [
                  - "sg-06fc404762de692db",
                  - "sg-0f04623ee5e2a8214",
                ]
              - self             = false
              - to_port          = 5432
            },
          - {
              - cidr_blocks      = []
              - description      = ""
              - from_port        = 5432
              - ipv6_cidr_blocks = []
              - prefix_list_ids  = []
              - protocol         = "tcp"
              - security_groups  = []
              - self             = false
              - to_port          = 5432
            },
          + {
              + cidr_blocks      = []
              + description      = null
              + from_port        = 5432
              + ipv6_cidr_blocks = []
              + prefix_list_ids  = []
              + protocol         = "tcp"
              + security_groups  = []
              + self             = false
              + to_port          = 5432
            },
            # (1 unchanged element hidden)
        ]
```

Hopefully if you merge `main` into this branch, that will clean that part up. And then, as discussed offline, make sure you create the bastion security group before merging this PR.

PR #280, opened 2021-07-02
Document PostGIS image & deprecated scraper

Mr0grog:
The reasons and use of these two components is not obviously clear to someone new to the codebase, so this adds some simple notices describing them.


astonm (approved):
👍🏽

PR #291, opened 2021-07-03
Update loader dependencies

Mr0grog:
This includes a couple major version updates. I've checked the release notes and tested in order to ensure things are compatible. It supersedes a few pull requests from Snyk (#284, #286).

The only package this does not update is Prettier, since it contains changes that would format our code differently (https://prettier.io/blog/2021/05/09/2.3.0.html). I’ll leave that for a later time, and we should probably do it across all projects at once.

### DIFF
--- a/loader/package-lock.json
+++ b/loader/package-lock.json
@@ -9,19 +9,19 @@
       "version": "0.1.0",
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


Mr0grog:
Accidentally committed out of sync lockfile.

PR #297, opened 2021-07-05
Don't output to STDOUT if sending via HTTP

### LINKED ISSUE #221: Loader logs are expensive
While working on #208, I noticed that about 40% of our costs in AWS are coming directly from log ingestion fees. These are mostly from the loaders, and probably due to the fact that they log every record they create:

Screen Shot 2021-06-16 at 12 33 27 AM

[Sum of bytes ingested by day per task — view in CloudWatch](https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#metricsV2:graph=~(metrics~(~(~'AWS*2fLogs~'IncomingBytes~'LogGroupName~'*2fecs*2fdaily-data-snapshot)~(~'...~'*2fecs*2fcvs-smart-api)~(~'...~'*2fecs*2fwaDoh)~(~'...~'*2fecs*2fnjvss)~(~'...~'*2fecs*2friteAidApi)~(~'...~'*2fecs*2fapi)~(~'...~'*2fecs*2fvaccinespotter))~legend~(position~'right)~region~'us-west-2~liveData~false~start~'-PT672H~end~'P0D~title~'IncomingBytes*3a*20Sum~view~'timeSeries~stacked~false~stat~'Sum~period~86400))

Should we consider not logging so much detail? These can be useful for debugging, but TBH I find myself mostly looking at error logs rather than the availability data that loaders sent. When I need the raw availability data that was sent, I almost always get something useful enough from the metadata in Sentry (which includes the POST body of updates that errored in the API server).

Mr0grog:
The loader currently streams output data on STDOUT, but when run on an automated schedule and using `--send` to send data to the API, that results in *huge* logfiles that aren't really useful for solving most problems and are instead very costly to manage. To avoid those issues, using the `--send` option now prevents data from being written to STDOUT.

This is a little easier than directing the output to `/dev/null` in Docker, since we’d need to add a bash script for that, and would then need `dumb-init` or another init system, and then things just get a bit complicated. Not writing to STDOUT when sending via HTTP seems like a reasonable switch, and keeps things simple.

(I initially tried to do something a little fancier by adding an option to specify a file path to write output to, but realized we don’t have any real use cases for it, so the added complexity didn’t make a whole lot of sense.)

Fixes #221.

### DIFF
--- a/loader/src/cli.js
+++ b/loader/src/cli.js
@@ -5,6 +5,7 @@ const Sentry = require("@sentry/node");
 const yargs = require("yargs");
 const { ApiClient } = require("./api-client");
 const { sources } = require("./index");
+const { oneLine } = require("./utils");
 
 Sentry.init();
 
@@ -44,19 +45,14 @@ function createDatabaseSender() {
   };
 }
 
-function compoundHandler(...handlers) {
-  return function handler(...args) {
-    handlers.forEach((handle) => handle(...args));
-  };
-}
-
 // Returns true on success, and false on failure.
 async function run(options) {
-  const jsonSpacing = options.compact ? 0 : 2;
-
-  let handler = createResultLogger(jsonSpacing);
+  let handler;
   if (options.send) {
-    handler = compoundHandler(handler, createDatabaseSender());
+    handler = createDatabaseSender();
+  } else {
+    const jsonSpacing = options.compact ? 0 : 2;
+    handler = createResultLogger(jsonSpacing);
   }
 
   let success = true;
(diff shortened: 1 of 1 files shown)


loader/src/cli.js
```diff
@@ -130,23 +126,31 @@ function main() {
         Load data about COVID-19 vaccine appointment availability from a
         variety of different sources.
 
+        Data about each vaccination location is written as a single line of JSON
+        on STDOUT, so you can pipe or stream the output to other files or
+        programs for processing. Informational messages are available on STDERR.
+
```
Mr0grog, line 132: This isn’t anything new, but it seemed like the current behavior might be worth stating more explicitly.

Mr0grog:
Going ahead and merging this as an important cost reduction. Will keep an eye on Sentry for any issues.

astonm (approved):
:shipit:

PR #298, opened 2021-07-05
Issue 223: Validate colon not present when saving provider_location.externalId

### LINKED ISSUE #223: Validate that external ID systems do not have colons
The question: “are we sure that no external ID systems have `:` in their name?” keeps coming up in various discussions. We have been pretty deliberate about not doing this, but that has also necessarily required being consciously careful. Instead, we should add validation that rejects any system names with colons.

The most straightforward thing to do here is probably to throw an exception in `db.addExternalIds()` if the system name has a colon here:

https://github.com/usdigitalresponse/appointment-availability-infra/blob/c7a66c5bfd092cf414432f7bc7dab0cec79a49c1/server/src/db.ts#L116-L140

And we should have at least one test for this, too.

channiemills:
This PR addresses #223 by throwing an error if `:` is present in the `system` component of a `provider_location.externalId`. Note, this does not check if `:` is present in `value`.  

If `:` is present on create, the provider location is created without any external ids. If present on update, the original external Ids are preserved.

Let me know if there are any thoughts or feedback here!

Fixes #223

### DIFF
--- a/server/src/db.ts
+++ b/server/src/db.ts
@@ -133,7 +133,7 @@ export async function addExternalIds(
       toInsert.map(([system, value]: [string, string]) => {
         if (system.includes(":")) {
           throw new ValueError(
-            `Provider location ${id} externalIDs include ${system} \
+            `Provider location ${id} externalIds include ${system} \
             but ':' is not allowed`
           );
         }
--- a/server/src/db.ts
+++ b/server/src/db.ts
@@ -134,7 +134,7 @@ export async function addExternalIds(
         if (system.includes(":")) {
           throw new ValueError(
             `Provider location ${id} externalIds include ${system} \
-            but ':' is not allowed`
+            but ':' is not allowed`.replace(/\n\s*/g, " ")
           );
         }
         return {
--- a/server/src/db.ts
+++ b/server/src/db.ts
@@ -131,11 +131,11 @@ export async function addExternalIds(
   await dbConn("external_ids")
     .insert(
       toInsert.map(([system, value]: [string, string]) => {
-        if (system.includes(':')) {
+        if (system.includes(":")) {
           throw new ValueError(
             `Provider location ${id} externalIDs include ${system} \
             but ':' is not allowed`
-          )
+          );
         }
         return {
           provider_location_id: id,
--- a/server/src/db.ts
+++ b/server/src/db.ts
@@ -131,6 +131,12 @@ export async function addExternalIds(
   await dbConn("external_ids")
     .insert(
       toInsert.map(([system, value]: [string, string]) => {
+        if (system.includes(':')) {
+          throw new ValueError(
+            `Provider location ${id} externalIDs include ${system} \
+            but ':' is not allowed`
+          )
+        }
         return {
           provider_location_id: id,
           system,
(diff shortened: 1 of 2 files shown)


server/src/db.ts
```diff
@@ -131,6 +131,12 @@ export async function addExternalIds(
   await dbConn("external_ids")
     .insert(
       toInsert.map(([system, value]: [string, string]) => {
+        if (system.includes(":")) {
+          throw new ValueError(
+            `Provider location ${id} externalIds include ${system} \
+            but ':' is not allowed`
```
Mr0grog, line 137: This still leaves an awkwardly large gap of several spaces between the system name and “but.” It might be worth porting `oneLine` over from the loader for this:

https://github.com/usdigitalresponse/univaf/blob/25a12f1cfb717a4bf5e57e53a0ddc8bd7d4e134c/loader/src/utils.js#L82-L108

Otherwise you could use `.replace(/\n\s*/g, " ")` instead:

```suggestion
            `Provider location ${id} externalIds include ${system}
            but ':' is not allowed`.replace(/\n\s*/g, " ")
```

(The difference with the template tag is that it doesn’t touch the content of the variables.)

server/test/db.test.ts
```diff
@@ -613,3 +614,43 @@ describe("db.getCurrentAvailabilityForLocations", () => {
...
+        ...updatedData,
+      };
+      await expect(async () => {
+        await createLocation(TestLocationInvalidExternalId);
+      }).rejects.toThrow(ValueError);
+
+      const { external_ids } = await getLocationById(
+        TestLocationInvalidExternalId.id
+      );
+      expect(external_ids).toEqual([]);
```
Mr0grog, line 639: > If `:` is present on create, the provider location is created without any external ids. If present on update, the original external Ids are preserved.

So that shouldn’t actually happen; the transaction wrapper here should cause the location not to be created at all: https://github.com/usdigitalresponse/univaf/blob/25a12f1cfb717a4bf5e57e53a0ddc8bd7d4e134c/server/src/db.ts#L102-L115

(And that seems to work when I test this PR manually.)

I’m pretty sure what’s happening in this test is that because the whole test itself is wrapped in a transaction ([that’s what `installTestDatabaseHooks()` does](https://github.com/usdigitalresponse/univaf/blob/25a12f1cfb717a4bf5e57e53a0ddc8bd7d4e134c/server/test/lib.ts#L65-L70)), the inner transaction in `createLocation` is effectively a no-op. I wonder if we need an alternative version of `installTestDatabaseHooks()` that truncates tables instead… not sure that’s critically necessary for this PR, though it would be nice.

Mr0grog (commented):
This works great! Left two thoughts inline about the error message formatting and the fact that our current testing approach doesn’t let you test the correct behavior of `createLocation()` here. They aren’t critical, but I think it would be good to do them before merging.

channiemills:
As discussed on slack, I'll defer the updates to the hook in another PR. I previously tried truncating after each test but I ran into database integrity errors I believe because I did not truncate all application tables.

Mr0grog (approved):
Looks good to me! :shipit: 

If you post a draft PR with the truncation code that wasn't working, we can work through the issues together.

PR #299, opened 2021-07-06
Improve SMART SL API, use new aggregator guidance

### LINKED ISSUE #117: Update SMART Scheduling Links API to use new aggregator guidance
I recently helped add a section to the SMART Scheduling Links spec with guidance for slot aggregators (like UNIVAF): https://github.com/smart-on-fhir/smart-scheduling-links/blob/master/specification.md#slot-aggregators

It contains some better (and hopefully now more universal) approaches some of the sticky situations we deal with poorly in the current implementation, and we should update what we publish to follow that guidance.

Mr0grog:
This vastly improves our implementation of the SMART Scheduling Links standard. It follows the new section with features designed for aggregators and also adds actual slots or daily capacity where we have them, now that we have top-level fields for that info.

- Fixes identifiers to use the new `external_ids` format.
- Simplifies location and schedule IDs.
- Adds last updated time to locations.
- Adds the `has-availability` extension on schedules instead of the wonky way we used to handle unknown availability.
- Adds `lastSourceSync` meta extension.
- Lists actual individual slots instead of one placeholder slot per location. If we don't know slots but do know daily capcity, that's used to create slots instead.
- Adds booking phone and deep link on slots.
- Adds capacity on slots.
- Includes available products and doses on each schedule.

This doesn’t create separate schedules for each product & dose combination, but we could do that in the future. It’s still technically valid, although the standard prefers that each product have a separate schedule if possible.

This also passes the two validators:
- https://infernotest.healthit.gov/community/
- https://github.com/lazau/scheduling-links-aggregator/tree/main/validator

This also adds docs about all the various particulars or caveats to our implementation.

Fixes #117.

### DIFF
--- a/server/src/smart-scheduling-routes.ts
+++ b/server/src/smart-scheduling-routes.ts
@@ -253,14 +253,19 @@ export async function listSchedules(
           }
         }
         if (provider.availability?.doses) {
-          const doseNumbers = provider.availability.doses.flatMap(
-            (dose: string) => DOSE_NUMBERS[dose] || []
+          const doseNumbers = new Set(
+            provider.availability.doses.flatMap(
+              (dose: string) => DOSE_NUMBERS[dose] || []
+            )
           );
-          for (const doseNumber of new Set(doseNumbers)) {
-            extension.push({
-              url: EXTENSION_VACCINE_DOSE,
-              valueInteger: doseNumber,
-            });
+          // Ensure extensions numerically sorted (`DOSE_NUMBERS.*` should be).
+          for (const doseNumber of DOSE_NUMBERS.all_doses) {
+            if (doseNumbers.has(doseNumber)) {
+              extension.push({
+                url: EXTENSION_VACCINE_DOSE,
+                valueInteger: doseNumber,
+              });
+            }
           }
         }
 
--- a/server/src/smart-scheduling-routes.ts
+++ b/server/src/smart-scheduling-routes.ts
@@ -10,7 +10,12 @@
 import { Request, Response } from "express";
 import { getHostUrl } from "./config";
 import * as db from "./db";
-import { Availability, LocationAvailability } from "./interfaces";
+import {
+  Availability,
+  LocationAvailability,
+  SlotRecord,
+  CapacityRecord,
+} from "./interfaces";
 import states from "./states.json";
 
 const BOOKING_DEEP_LINK =
@@ -246,7 +251,7 @@ export async function listSchedules(
         }
         if (provider.availability?.doses) {
           const doseNumbers = provider.availability.doses.flatMap(
-            (dose) => DOSE_NUMBERS[dose] || []
+            (dose: string) => DOSE_NUMBERS[dose] || []
           );
           for (const doseNumber of new Set(doseNumbers)) {
             extension.push({
@@ -313,7 +318,7 @@ export async function listSlots(req: Request, res: Response): Promise<void> {
           provider.availability?.slots || provider.availability?.capacity;
         if (!slots || !slots.length) return [];
 
-        return slots.map((slot) => {
+        return slots.map((slot: SlotRecord | CapacityRecord) => {
           const extension: Array<any> = [];
 
           if (slot.available_count != null) {
--- a/server/src/smart-scheduling-routes.ts
+++ b/server/src/smart-scheduling-routes.ts
@@ -8,6 +8,7 @@
  */
 
 import { Request, Response } from "express";
+import { DateTime } from "luxon";
 import { getHostUrl } from "./config";
 import * as db from "./db";
 import {
@@ -64,6 +65,8 @@ const DOSE_NUMBERS: { [index: string]: number[] } = {
   first_dose_only: [1],
   second_dose_only: [2],
 };
+
+const INVALID_ID_CHARACTERS = /[^A-Z\d.-]/gi;
 
 interface FhirIssue {
   severity: "fatal" | "error" | "warning" | "information";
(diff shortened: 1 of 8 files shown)


server/src/smart-scheduling-routes.ts
```diff
@@ -176,10 +228,42 @@ export async function listSchedules(
...
+                display: PRODUCT_NAMES[product],
+              },
+            });
+          }
+        }
+        if (provider.availability?.doses) {
+          const doseNumbers = provider.availability.doses.flatMap(
+            (dose: string) => DOSE_NUMBERS[dose] || []
+          );
+          for (const doseNumber of new Set(doseNumbers)) {
```
astonm, line 256: nit: It would be nice (for some definition of "nice") to ensure that the dose numbers are sorted here. Not sure it's worth messing with given the significance of the rest of the diff, but if you do, it might be simpler just to loop through `DOSE_NUMBERS.all_doses` (which will presumably stay ordered) and check if each of the values is in this set.

server/src/smart-scheduling-routes.ts
```diff
@@ -218,46 +313,70 @@ export async function listSlots(req: Request, res: Response): Promise<void> {
...
+            end =
+              slot.end ||
+              new Date(
+                new Date(slot.start).getTime() + 15 * 60 * 1000
+              ).toISOString();
+            id = `${id}-${new Date(slot.start).getTime()}`;
+          } else {
+            start = `${slot.date}T00:00:00Z`;
+            end = `${slot.date}T00:00:00Z`;
+            id = `${id}-${slot.date}`;
```
astonm, line 358: Not sure whether it would buy you much here, but it might be good to pull out the time formatting stuff here into functions (or rely on `luxon`?)

server/test/smart-scheduling.test.ts
```diff
@@ -70,23 +71,174 @@ describe("GET /smart-scheduling/schedules/states/:state.ndjson", () => {
...
+    location = await createLocation({
+      ...TestLocation,
+      id: "123",
+      external_ids: [["random", "value"]],
+    });
+    await updateAvailability(location.id, {
+      ...TestLocation.availability,
+      available: Availability.UNKNOWN,
+    });
+    // This location has not availability record at all.
```
astonm, line 98:
```suggestion
    // This location has no availability record at all.
```

server/test/smart-scheduling.test.ts
```diff
@@ -70,23 +71,174 @@ describe("GET /smart-scheduling/schedules/states/:state.ndjson", () => {
     const data = ndjsonParse(res.body);
     expect(data).toHaveLength(1);
   });
+
+  it("shows available with `has-availability` extension", async () => {
+    let location = await createLocation(TestLocation);
+    await updateAvailability(location.id, {
```
astonm, line 77: nit: For the sake of clarity, maybe create all of these locations (and updating their availability) by looping over an array?

astonm (approved):
Looks pretty good to me. Just a few thoughts for you below.

server/src/smart-scheduling-routes.ts
Mr0grog: Can do that. I think this would also work, though it’s not very fun to read:

```suggestion
          for (const doseNumber of [...new Set(doseNumbers)].sort()) {
```

server/src/smart-scheduling-routes.ts
Mr0grog: Yeah, I guess I really should have done that for the math above. Not sure Luxon is super valuable to these particular lines since they’re so simple, but I hear you on cleaning this up a bit.

server/test/smart-scheduling.test.ts
Mr0grog: 813bd079b088cb5316d2cab7b58b5d60cb7d6154 is indeed much nicer. I should make use of `it.each()` more. :)

astonm: Ah yeah, very nice!

server/test/smart-scheduling.test.ts
```diff
@@ -1,15 +1,44 @@
...
+
+  if (customizations.availability) {
+    await updateAvailability(location.id, {
+      ...TestLocation.availability,
+      ...customizations.availability,
+    });
+  }
+
+  return location;
+}
```
Mr0grog, line 40: Might be worth moving to `test/lib` in the future? This was super useful for the `has-availability` tests, and I think it makes for a *slightly* smoother developer experience in tests that need both a location and availability, but it might not be a huge difference. 🤷

server/test/smart-scheduling.test.ts
```diff
@@ -70,23 +99,144 @@ describe("GET /smart-scheduling/schedules/states/:state.ndjson", () => {
...
+      // Test no availability record at all.
+      [null, "unknown"],
+    ])("represents %j as %j", async (available, expected) => {
+      const location = await createRandomLocation({
+        availability: available && { available },
+      });
+      const response = await context.client<any[]>(
+        `smart-scheduling/schedules/states/${location.state}.ndjson`,
+        { parseJson: ndjsonParse }
+      );
```
Mr0grog, line 122: Realized I could use the `parseJson` option to make it handle ND-JSON, which is neat. Wondering if `useServerForTests` should set up something like:

```js
context.client = got.extend({
  // ...currently existing options...

  parseJson (text: string) {
    // Unfortunately I don’t think we can get headers to switch on here, so just try both. :\
    try {
      return JSON.parse(text);
    } catch (_) {
      return ndjsonParse(text);
    }
  }
}
```

Would make other tests that need ND-JSON support nicer, too. And all our endpoints that return ND-JSON may also return regular JSON when there’s an error.

Mr0grog: Ha, now that I wrote that, I realize it won’t work well for ND-JSON responses with only one entry, since it’ll see them as normal JSON and you won’t get an array. 😞

astonm: Hm, yeah. Although (and don't do this!) you could always do something like this...

```javascript
parseJson (text: string) {
  try {
    const parsed = JSON.parse(text);
    const out = [parsed];
    Object.assign(out, parsed);
  } catch (_) {
    return ndjsonParse(text);
  }
}
```
...to get the effect of both worlds, so long as the resulting parsed object doesn't have any numerical keys. 😝

Mr0grog: 😳

Issue #300, opened 2021-07-07
Add data loader for CDC’s Socrata Data Portal


Mr0grog:
CDC is now publishing supply information (the same dataset that powers vaccines.gov) on their data portal at https://data.cdc.gov/Vaccinations/Vaccines-gov-COVID-19-vaccinating-provider-locatio/5jp2-pgaw. (This is a Socrata data portal, so can be accessed using Socrata’s standard API.)

This isn’t availability data, but does contain useful information about:
- What vaccine products are being supplied to a location (useful if we can’t get availability data about products from other sources).
- Other location-specific info, like location, position, hours, walk-ins allowed, etc.
- Vaccines.gov GUIDs that we can put in `external_ids`.

Each line is a unique location (vaccines.gov GUID) + vaccine product combination, so many locations will be listed more than once. Vaccine products are identified by their Unit-of-Use NDC10 code (the `ndc` column), which can be mapped to CVX codes and vaccine names with data from this CDC page: https://www.cdc.gov/vaccines/programs/iis/COVID-19-related-codes.html

This stuff is super useful (*especially* the product info), and we should write a loader that pulls it in regularly. It doesn’t update rapidly, so that loader should probably just run a couple times a day.

---

**This will also need some up-front work to map location GUIDs to our locations so we don’t create lots of duplicates the first time we run it.**
- The store number (`loc_store_no`) column might be useful for this, but requires having a way to identify which store brand that store number goes with.
- Vaccinate the States should also have concordances with these identifiers, and that might be a better/easier way to match things up initially.

Issue #212, opened 2021-06-15
VaccinateTheStates to UNIVAF keys


janovergoor:
VtS sometimes maps more than one location to a single one of ours:

```
id     uuid                             name                         address                                        zip   city            county     state
lghrr  05e0c103-3bdc-4a34-bd95-a139c48… The Little Clinic #6125      6000 S Holly St                                80111 Greenwood Vill… Arapahoe   CO   
lgwrc  05e0c103-3bdc-4a34-bd95-a139c48… King Soopers Pharmacy #6200… 6000 S Holly St                                80111 Greenwood Vill… Arapahoe   CO   
lqgzp  1b960a3a-54a1-409b-99fa-7dd374f… Walmart Inc #10-3176         1725 N Union Blvd                              -2850 Colorado Sprin… El Paso    CO   
lcdtqb 1b960a3a-54a1-409b-99fa-7dd374f… Tesla High School - Nomi He… 2111 Carlton AveColorado Springs, CO 80909     80909 Colorado Sprin… El Paso    CO   
ldxdz  ffee50a8-6d4f-42b3-8ced-d9fceb1… Osco Drug #3262              4660 W. Irving Park Road                       60641 Chicago         Cook       IL   
lmprg  ffee50a8-6d4f-42b3-8ced-d9fceb1… Walgreens Co. #4936          4040 N Cicero Ave                              -1807 Chicago         Cook       IL   
lgchf  5e3eaa28-16b0-4ac7-aa64-00c2e77… Central Pharmacy-Laingsburg  209 E Grand River Rd                           48848 Laingsburg      Shiawassee MI   
lcdrxc 5e3eaa28-16b0-4ac7-aa64-00c2e77… Central Pharmacy Laingsburg  7400 WOODBURY ROAD LAINGSBURG MILaingsburg, M… 48848 Laingsburg      Shiawassee MI   
lxwcf  54ff60f9-0c5f-4b13-920c-c245fcf… MEDICINE SHOPPE PHARMACY     149 POND FORT TRLLAKE ST LOUIS, MO 63367       63367 LAKE ST LOUIS   St. Charl… MO   
lcdtry 54ff60f9-0c5f-4b13-920c-c245fcf… CVS #18053                   7909 Hwy NO'fallon, MO 63368                   63368 O'fallon        St. Charl… MO   
[...]

Issue #123, opened 2021-05-13
Database setup script no longer works after UUID changes


astonm:
This is the error message I get running `scripts/build-database.sh`:

```
error: SELECT id, location_id, source
    FROM availability
    WHERE location_id = $1 AND source = $2 - invalid input syntax for type uuid: "47c59c23cbd4672173cc93b8a39b60ddf481dd56"
    at Parser.parseErrorMessage (/Users/aston/usdr/appointment-availability-infra/server/node_modules/pg-protocol/dist/parser.js:278:15)
    at Parser.handlePacket (/Users/aston/usdr/appointment-availability-infra/server/node_modules/pg-protocol/dist/parser.js:126:29)
    at Parser.parse (/Users/aston/usdr/appointment-availability-infra/server/node_modules/pg-protocol/dist/parser.js:39:38)
    at Socket.<anonymous> (/Users/aston/usdr/appointment-availability-infra/server/node_modules/pg-protocol/dist/index.js:10:42)
    at Socket.emit (node:events:365:28)
    at addChunk (node:internal/streams/readable:314:12)
    at readableAddChunk (node:internal/streams/readable:289:9)
    at Socket.Readable.push (node:internal/streams/readable:228:10)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23) {
  length: 135,
  severity: 'ERROR',
  code: '22P02',
  detail: undefined,
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'uuid.c',
  line: '137',
  routine: 'string_to_uuid'
}'
```

It appears to fail inside `updateAvailability`.

PR #304, opened 2021-07-13
Load geographic coordinates for Rite Aid and CVS

### LINKED ISSUE #236: Get geographic coordinates for Rite Aid and CVS
Our lack of geographic coordinates for most Rite Aid pharmacies (2,426 of 2,441 locations) and all CVS pharmacies (9,781) is causing some matching issues for Vaccinate the States. We should prioritize getting positions for these locations.

In many (maybe all?) instances, VtS already has these locations, and we can identify them with VtS’s concordances and our external_ids. 

The easiest way to do this right now might be to implement a very narrow version of #189 that just handles Rite Aid and CVS locations and run it once.

astonm:
This PR sets up a new loader module `vts.geo` intended to update our geolocation data from Vaccinate the States's API. For the sake of more narrowly solving #236, it only works for Rite Aid and CVS locations, however I don't know of a reason why it wouldn't work for everything.

In addition to updating the `provider_location.position` field, it also may update `provider_location.name` if there is a mismatch. Most of our names *don't* match, but I did some regular expression hacking to try to get easy ones to align. To be honest, I wish it didn't touch the names, but the server requires sending one.

I haven't written any automated tests, yet. I'm not sure what would be best to test here. I did manually test this somewhat extensively, though. For the most part it does the right thing. However, I ran into a few surprises in the data set that cause issues.

First, there are some concordance values that are not unique to a location. The easy one to exclude was `_tag_provider`, which e.g. has the value `cvs` for every CVS location. But there were others that I noticed were not so reliable, like `google_places` and `placekey`. I opted to pull a shortlist of `cvs`,`rite_aid`,`vaccinefinder_org` and `vaccinespotter_org`.

Second, even with that degree of filtering, there are still some erroneous values in the data set, e.g. 

*(from https://api.vaccinatethestates.com/v0/CA.geojson)*
```json
{
  "name": "RITE AID PHARMACY 05435",
  "concordances": [
    "google_places:ChIJRzqoPqzAwoAR8AB9GQTJ6-w",
    "vaccinefinder_org:4f3cea2e-7547-4997-abf0-668e083d7f9c",
    "placekey:zzw-22f@5z5-3nz-tqf",
    "rite_aid:105435",
    "rite_aid:5436",
    "_tag_provider:rite_aid",
    "us_carbon_health:584b4f6a-0c58-4580-9354-04777cf9ca3d",
    "us_carbon_health:86c9c4fe-df2c-413c-9f9e-cf619bbe4600",
    "us_carbon_health:a515f425-e50d-4292-be59-966ae4924100",
    "vaccinespotter_org:7528795"
  ]
}
```
Which has `rite_aid:105435` and `rite_aid:5436` (the latter seems to be a typo, and there's another `rite_aid:5436` for real in the data set).

The consequence of these buggy concordances mostly mean that we may write wrong names and locations, but I'm not sure how best to handle them.

Fixes #236

### DIFF
--- a/loader/src/sources/vts/geo.js
+++ b/loader/src/sources/vts/geo.js
@@ -60,6 +60,38 @@ function hasUsefulData(store) {
   );
 }
 
+function splitConcordance(concordance) {
+  const colon = concordance.indexOf(":");
+  return [concordance.substring(0, colon), concordance.substring(colon + 1)];
+}
+
+const systemsToSend = {
+  cvs: "cvs",
+  rite_aid: "rite_aid",
+  vaccinespotter_org: "vaccinespotter",
+  vaccinefinder_org: "vaccines_gov",
+  vaccinefinder: "vaccines_gov",
+};
+
+function renameSystem([systemCode, value]) {
+  const system = systemsToSend[systemCode];
+  if (!system) {
+    throw new Error(`Unexpected system code ${systemCode}`);
+  }
+  return [system, value];
+}
+
+function validConcordance([systemCode, value]) {
+  const system = systemsToSend[systemCode];
+  if (!system) {
+    return false;
+  }
+  if (system == "rite_aid" && value.match(/^1\d{5}$/)) {
+    return false;
+  }
+  return true;
+}
+
 function formatStore(store) {
   const data = store.properties;
 
@@ -81,43 +113,8 @@ function formatStore(store) {
       provider: provider,
     });
 
-    function splitConcordance(concordance) {
-      const colon = concordance.indexOf(":");
-      return [
-        concordance.substring(0, colon),
-        concordance.substring(colon + 1),
-      ];
-    }
-
-    function filterConcordancePairs(pairs) {
-      return pairs
-        .filter(([systemCode, value]) => {
-          const system = systemsToSend[systemCode];
-          if (!system) {
-            return null;
-          }
-
-          if (system == "rite_aid" && value.match(/^1\d{5}$/)) {
-            // Stores like Rite Aid in this data set may have a store number like 105612.
-            // IDs with that shape (6 digits, starting with '1') are likely misparsed from VTrckS.
-            return null;
-          }
-
-          return [system, value];
-        })
-        .filter(Boolean);
-    }
-
-    const systemsToSend = {
-      cvs: "cvs",
-      rite_aid: "rite_aid",
-      vaccinespotter_org: "vaccinespotter",
-      vaccinefinder_org: "vaccines_gov",
-      vaccinefinder: "vaccines_gov",
-    };
-
     const concordances = data.concordances.map(splitConcordance);
-    const externalIds = filterConcordancePairs(concordances);
+    const externalIds = concordances.filter(validConcordance).map(renameSystem);
     const univafPair = concordances.filter((v) => v[0] == "getmyvax_org")[0];
 
     if (!externalIds.length) {
--- a/loader/src/sources/vts/geo.js
+++ b/loader/src/sources/vts/geo.js
@@ -99,9 +99,9 @@ function formatStore(store) {
       "vaccinespotter_org",
     ]);
 
-    const externalIds = data.concordances
-      .map(splitConcordance)
-      .filter((v) => systemsToSend.has(v[0]));
+    const concordances = data.concordances.map(splitConcordance);
+    const externalIds = concordances.filter((v) => systemsToSend.has(v[0]));
+    const univafPair = concordances.filter((v) => v[0] == "getmyvax_org")[0];
 
     if (!externalIds.length) {
       return null;
@@ -110,6 +110,7 @@ function formatStore(store) {
     result = {
       name,
       provider,
+      id: univafPair && univafPair[1],
       external_ids: externalIds,
       position: {
         longitude: store.geometry.coordinates[0],
--- a/loader/src/sources/vts/geo.js
+++ b/loader/src/sources/vts/geo.js
@@ -92,9 +92,20 @@ function formatStore(store) {
       ];
     }
 
+    systemsToSend = new Set([
+      "cvs",
+      "rite_aid",
+      "vaccinefinder_org",
+      "vaccinespotter_org",
+    ]);
+
     const externalIds = data.concordances
       .map(splitConcordance)
-      .filter((v) => v[0] != "getmyvax_org"); // ignore getmyvax concordances
+      .filter((v) => systemsToSend.has(v[0]));
+
+    if (!externalIds.length) {
+      return null;
+    }
 
     result = {
       name,
@@ -122,6 +133,7 @@ async function updateGeo(handler, options) {
     const formatted = stores.features
       .filter(hasUsefulData)
       .map(formatStore)
+      .filter(Boolean)
       .forEach((item) => handler(item, { update_location: true }));
 
     results = results.concat(formatted);
--- a/loader/src/sources/vts/geo.js
+++ b/loader/src/sources/vts/geo.js
@@ -143,6 +143,7 @@ async function updateGeo(handler, options) {
 
   if (!states || !states.length) {
     warn("No states specified for vts.geo");
+    return [];
   }
 
   const statesFilter = new Set(states);
--- a/loader/src/sources/vts/geo.js
+++ b/loader/src/sources/vts/geo.js
@@ -3,7 +3,29 @@ const Sentry = require("@sentry/node");
 const { Available, LocationType } = require("../../model");
 const { titleCase } = require("../../utils");
 
-updateworthyProviders = new Set(["CVS", "Rite-Aid"]);
+dataProviders = {
+  "Rite-Aid": {
+    provider: "rite_aid",
+    getStoreName: (data) => {
+      const m = data.name.match(/RITE AID PHARMACY #?(\d+)/i);
+      if (m) {
+        return `Rite Aid #${parseInt(m[1], 10)}`; // format as in riteaid/api.js
+      }
+      return data.name;
+    },
+  },
+  CVS: {
+    // modeled after code in cvs/api.js
+    provider: "cvs",
+    getStoreName: (data) => {
+      const m = data.name.match(/CVS Pharmacy, Inc. #?(\d+)/i);
+      if (m) {
+        return `CVS #${parseInt(m[1], 10)}`; // format as in cvs/api.js
+      }
+      return data.name;
+    },
+  },
+};
 
 function warn(message, context) {
   console.warn(`VTS Geo: ${message}`, context);
@@ -36,7 +58,8 @@ function hasUsefulData(store) {
   return (
     store.properties.concordances?.length &&
     store.geometry?.coordinates &&
-    updateworthyProviders.has(store.properties.provider?.name)
+    store.properties.provider?.name &&
+    store.properties.provider?.name in dataProviders
   );
 }
 
@@ -45,11 +68,20 @@ function formatStore(store) {
 
   let result;
   Sentry.withScope((scope) => {
+    const dataProvider = dataProviders[data.provider.name];
+    if (!dataProvider) {
+      error(`Unexpected provider name ${data.provider.name}`);
+      return null;
+    }
+
+    const provider = dataProvider.provider;
+    const name = dataProvider.getStoreName(data);
+
     scope.setContext("location", {
       id: store.id,
       url: data.vts_url,
-      name: data.name,
-      provider: data.provider,
+      name: name,
+      provider: provider,
     });
 
     function splitConcordance(concordance) {
@@ -65,7 +97,8 @@ function formatStore(store) {
       .filter((v) => v[0] != "getmyvax_org"); // ignore getmyvax concordances
 
     result = {
-      name: data.name,
+      name,
+      provider,
       external_ids: externalIds,
       position: {
         longitude: store.geometry.coordinates[0],
--- a/loader/src/sources/vts/geo.js
+++ b/loader/src/sources/vts/geo.js
@@ -111,7 +111,6 @@ function formatStore(store) {
 
     const concordances = data.concordances.map((c) => splitOnce(c, ":"));
     const externalIds = concordances.filter(validConcordance).map(renameSystem);
-    const univafPair = concordances.filter((v) => v[0] == "getmyvax_org")[0];
 
     if (!externalIds.length) {
       return null;
@@ -120,7 +119,6 @@ function formatStore(store) {
     result = {
       name,
       provider,
-      id: univafPair ? univafPair[1] : undefined,
       external_ids: externalIds,
       position: {
         longitude: store.geometry.coordinates[0],
--- a/loader/src/sources/vts/geo.js
+++ b/loader/src/sources/vts/geo.js
@@ -41,8 +41,8 @@ async function queryState(state) {
       url: `https://api.vaccinatethestates.com/v0/${state}.geojson`,
     });
     return JSON.parse(response.body);
-  } catch (error) {
-    error(`Error fetching Vaccine Spotter data`, error);
+  } catch (e) {
+    error(`Error fetching Vaccine Spotter data`, e);
     return [];
   }
 }
--- a/loader/src/sources/vts/geo.js
+++ b/loader/src/sources/vts/geo.js
@@ -35,14 +35,14 @@ function error(message, context) {
   Sentry.captureMessage(message, Sentry.Severity.Error);
 }
 
-async function queryState(state) {
+async function getStores(state) {
   try {
     const response = await got({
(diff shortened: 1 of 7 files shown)


loader/src/sources/vts/geo.js
```diff
@@ -0,0 +1,146 @@
...
+
+    const systemsToSend = new Set([
+      "cvs",
+      "rite_aid",
+      "vaccinefinder_org",
+      "vaccinespotter_org",
+    ]);
+
+    const concordances = data.concordances.map(splitConcordance);
+    const externalIds = concordances.filter((v) => systemsToSend.has(v[0]));
```
Mr0grog, line 101: We need a mechanism for remapping names here — in our external IDs, `vaccinespotter_org` is just `vaccinespotter` (e.g. http://getmyvax.org/api/edge/locations/vaccinespotter:141533984).

(While we’re at it, I wonder if we should rename `vaccinefinder_org` to `vaccines_gov`? Probably not a huge deal.)

Mr0grog (commented):
Ughhhh, this is way less clean than I had expected. Thanks for running into and calling out some of these issues.

In your example, you had a location with `rite_aid:105435` and `rite_aid:5436`. The first one isn’t the store number, even though it’s similar (I’m just about certain this is the federal-level VTrckS PIN minus the “RA” prefix — Rite Aid’s and most national pharamcy chains’ federal-level VTrckS PINs are based on their store numbers like this), so it won’t match up for us. And as you noted the second one is just wrong, and identifies this as a *different* store (which does not actually provide vaccinations). Both of these seem really problematic.

Thought: since we know we have a comprehensive list of locations for CVS and Rite Aid, what if we started by pulling *our* list of locations, then matching them up with the data from VtS by the `getmyvax_org` concordance if present, otherwise the `vaccinespotter_org` concordance (which is just `vaccinespotter` for us), and otherwise by parsing the ID out of the name? (And not including a record if we can’t match it by one of those means.) That’s a little messy, but feels more likely to get a good match to me.

Otherwise, maybe we should use CDC’s new open data for this instead. (#300)

---

Other messy sidenote here: we use an unpadded number for Rite Aid external IDs: https://github.com/usdigitalresponse/univaf/blob/bcf5aca4e9b169723843478177973efa12bebc18/loader/src/sources/riteaid/api.js#L100

But we use a 5-digit 0-padded number for CVS (because legacy): https://github.com/usdigitalresponse/univaf/blob/bcf5aca4e9b169723843478177973efa12bebc18/loader/src/sources/cvs/smart.js#L183-L185

loader/src/sources/vts/geo.js
Mr0grog: Also realized some locations use `vaccinefinder_org` while others use `vaccinefinder` as the concordance, and some even use both.

astonm:
@Mr0grog with the changes in #309 I think we should get reliable matches for CVS in this data se, and my latest commit addresses the renaming you mentioned (let me know if I didn't get it right or if we should do others, too).

I'm not able to find cases that use ` vaccinefinder` though. I think everything on their side uses ` vaccinefinder_org`?

astonm:
> Here’s one from New Jersey

Ah, thanks. I had a (big) bug in my script to pull them all. I'll add that to the list of transformed systems.

Anything else before you think this is ready to go?

astonm:
@Mr0grog I just repointed this script at the S3 dump, and it might be ready to go.

One thing I noticed, there are entries that look like this in the data set:

```json
  "external_ids": [
    ...
    [
      "rite_aid",
      "105612"
    ],
    [
      "rite_aid",
      "5612"
    ],
    ...
  ],
```

That `105612` store ID looks suspiciously like the special handling of the leading `10` for Rite Aid we saw in the VTrckS IDs from #312. Do you think it's worth cleaning them up (e.g. maybe by only using the shortest identifier for a given external ID system)?

Mr0grog:
Huh, I wonder how those got in there originally. It does seem like it would be good to clean them up, but I’m worried “shortest identifier for a given ID system” is a safe rule here. It might be smarter to go case-by-case for these, e.g. for `rite_aid`, we can look for 6 characters and starts with a `1`.

Sorry I wasn’t able to get to this earlier — was massively ill Friday and Saturday. :(

astonm:
@Mr0grog Cool, filtering out the Rite Aid IDs that are shaped as you described. It seems like all of the remaining ones (from Rite Aid and from CVS) are good.

loader/src/sources/vts/geo.js
```diff
@@ -0,0 +1,164 @@
...
+      cvs: "cvs",
+      rite_aid: "rite_aid",
+      vaccinespotter_org: "vaccinespotter",
+      vaccinefinder_org: "vaccines_gov",
+      vaccinefinder: "vaccines_gov",
+    };
+
+    const concordances = data.concordances.map(splitConcordance);
+    const externalIds = filterConcordancePairs(concordances);
+    const univafPair = concordances.filter((v) => v[0] == "getmyvax_org")[0];
```
jaronoff97, line 122: Are we always guaranteed to have at least one element after the filter?

jaronoff97 (approved):

loader/src/sources/vts/geo.js
```diff
@@ -0,0 +1,164 @@
...
+    provider: "rite_aid",
+    getStoreName: (data) => {
+      const m = data.name.match(/RITE AID PHARMACY #?(\d+)/i);
+      if (m) {
+        return `Rite Aid #${parseInt(m[1], 10)}`; // format as in riteaid/api.js
+      }
+      return data.name;
+    },
+  },
+  CVS: {
```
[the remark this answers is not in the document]
astonm: No it doesn't, but I'll tweak for the sake of visual consistency.

loader/src/sources/vts/geo.js
```diff
@@ -0,0 +1,164 @@
+const got = require("got");
+const Sentry = require("@sentry/node");
+
+const dataProviders = {
+  "Rite-Aid": {
+    provider: "rite_aid",
+    getStoreName: (data) => {
+      const m = data.name.match(/RITE AID PHARMACY #?(\d+)/i);
+      if (m) {
+        return `Rite Aid #${parseInt(m[1], 10)}`; // format as in riteaid/api.js
```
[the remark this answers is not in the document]
astonm: The full string that matched the regex (e.g. `"Rite Aid Pharmacy #123 in Tuscaloosa, AL"`)

loader/src/sources/vts/geo.js
```diff
@@ -0,0 +1,164 @@
...
+
+function error(message, context) {
+  console.error(`VTS Geo: ${message}`, context);
+  Sentry.captureMessage(message, Sentry.Severity.Error);
+}
+
+async function getStores() {
+  try {
+    const response = await got({
+      url: `https://univaf-data-snapshots.s3.us-west-2.amazonaws.com/vts/vts-final-output-locations.geojson`,
```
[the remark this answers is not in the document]
astonm: See https://geojson.org/ for the broad spec, or you can just look at the data itself 😄

loader/src/sources/vts/geo.js
```diff
@@ -0,0 +1,164 @@
...
+      provider: provider,
+    });
+
+    function splitConcordance(concordance) {
+      const colon = concordance.indexOf(":");
+      return [
+        concordance.substring(0, colon),
+        concordance.substring(colon + 1),
+      ];
+    }
```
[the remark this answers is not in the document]
astonm: I probably prefer to inline such things, but it's just a little too long for that, so I made it a local function. I considered pulling it out completely, but it's only used in this once place, and the name doesn't exactly help you guess what it does out of context, so I wanted to keep it close to where it's invoked.

loader/src/sources/vts/geo.js
```diff
@@ -0,0 +1,164 @@
...
+
+    function filterConcordancePairs(pairs) {
+      return pairs
+        .filter(([systemCode, value]) => {
+          const system = systemsToSend[systemCode];
```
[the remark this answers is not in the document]
astonm: It looks funny, but does not matter since the name isn't looked up until later when it's called. Arguably there's too much all in this area, but I was trying to do it within the `Sentry.withScope` and `scope.setContext`.

loader/src/sources/vts/geo.js
astonm: We are not! If there isn't a concordance for `getmyvax_org` in the list, `univafPair` will end up `undefined`, which we guard against below.

loader/src/sources/vts/geo.js
```diff
@@ -0,0 +1,164 @@
...
+    const univafPair = concordances.filter((v) => v[0] == "getmyvax_org")[0];
+
+    if (!externalIds.length) {
+      return null;
+    }
+
+    result = {
+      name,
+      provider,
+      id: univafPair ? univafPair[1] : undefined,
```
Mr0grog, line 131: I feel a little iffy about actually using this mapping because of issues like #212. Did you do any checks to see how good a match this is likely to be or how often a location has more than one of our IDs or one of our IDs is used across multiple locations?

loader/src/sources/vts/geo.js
```diff
@@ -0,0 +1,164 @@
...
+    };
+  });
+  return result;
+}
+
+async function updateGeo(handler, options) {
+  const states = options.states?.split(",").map((state) => state.trim());
+
+  if (!states || !states.length) {
+    warn("No states specified for vts.geo");
```
Mr0grog, line 146: This should throw or return.

Not sending a second argument also leads to some funky looking output:

```
VTS Geo: No states specified for vts.geo undefined
```

Mr0grog (changes requested):
Mostly minor nits, but one major issue where concordance names aren’t actually being changed to the correct external ID systems.

Mr0grog:
This could probably also use at least some very simple tests.

loader/src/sources/vts/geo.js
```diff
@@ -0,0 +1,164 @@
...
+      const colon = concordance.indexOf(":");
+      return [
+        concordance.substring(0, colon),
+        concordance.substring(colon + 1),
+      ];
+    }
+
+    function filterConcordancePairs(pairs) {
+      return pairs
+        .filter(([systemCode, value]) => {
```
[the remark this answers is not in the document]
astonm: Oof, missed this somehow, thanks. It was originally intended to work differently and did actually filter, hence the name. Will clean this up.

loader/src/sources/vts/geo.js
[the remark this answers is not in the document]
astonm: Do you mean this literally, or a more generalized version that takes the delimiter as an argument?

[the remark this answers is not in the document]
Mr0grog: Sorry, a more generalized one that takes a delimiter (e.g. in FHIR/SMART SL references, the delimiter is `/`).

astonm:
@Mr0grog thanks. Made some updates and will add some tests soon. Other than the utility functions, any particular things you'd want to make sure we test here?

Mr0grog:
I think the most essential is just having a handful of real-world inputs and checking them against an expected output is the core thing (that would catch regressions like the external IDs). I think I’m OK if that's as deep as testing goes here for now, and don't have major concerns about any particular part needing deeper testing.

loader/src/sources/vts/geo.js
astonm: In this data set there's only one duplicated `getmyvax_org` id. Here are the stores:

```json
{
  "type": "Feature",
  "id": "lxwcf",
  "properties": {
    "name": "MEDICINE SHOPPE PHARMACY",
    "provider": {
      "name": "CVS",
      "provider_type": "Pharmacy",
      "vaccine_info_url": "https://www.cvs.com/immunizations/covid-19-vaccine"
    },
    "state": "MO",
    "location_type": "Unknown",
    "phone_number": null,
    "full_address": "149 POND FORT TRL\nLAKE ST LOUIS, MO 63367",
    "city": "LAKE ST LOUIS",
    "county": "St. Charles",
    "zip_code": "63367",
    "hours": {
      "unstructured": null,
      "structured": null
    },
    "website": null,
    "vaccines_offered": null,
    "accepts_appointments": null,
    "accepts_walkins": null,
    "concordances": [
      "mo_arcgis:2a9658e6-f87c-4299-929d-cccaba3b879e",
      "placekey:222-222@5pk-qs2-4gk",
      "_tag_provider:cvs",
      "placekey:zzw-224@5pk-qrx-6c5",
      "vtrcks:CV1018053",
      "getmyvax_org:54ff60f9-0c5f-4b13-920c-c245fcf6424c",
      "cvs:18053"
    ],
    "last_verified_by_vts": null,
    "vts_url": "https://www.vaccinatethestates.com/?lng=-90.78767&lat=38.76361#lxwcf"
  },
  "geometry": {
    "type": "Point",
    "coordinates": [
      -90.78767,
      38.76361
    ]
  }
}
{
  "type": "Feature",
  "id": "lcdtry",
  "properties": {
    "name": "CVS #18053 (inside Schnuck's)",
    "provider": {
      "name": "CVS",
      "provider_type": "Pharmacy",
      "vaccine_info_url": "https://www.cvs.com/immunizations/covid-19-vaccine"
    },
    "state": "MO",
    "location_type": "Pharmacy",
    "phone_number": "(888) 607-4287",
    "full_address": "7909 Highway N. Dardenne Prairie, MO 63368",
    "city": "Dardenne Prairie",
    "county": "St. Charles",
    "zip_code": "63368",
    "hours": {
      "unstructured": "",
      "structured": null
    },
    "website": "https://www.cvs.com/vaccine/intake/store/cvd-schedule",
    "vaccines_offered": null,
    "accepts_appointments": true,
    "accepts_walkins": false,
    "concordances": [
      "_tag_provider:cvs",
      "placekey:zzw-224@5pk-qrx-6c5",
      "vtrcks:CV1018053",
      "getmyvax_org:54ff60f9-0c5f-4b13-920c-c245fcf6424c",
      "cvs:18053"
    ],
    "last_verified_by_vts": null,
    "vts_url": "https://www.vaccinatethestates.com/?lng=-90.76839&lat=38.76884#lcdtry"
  },
  "geometry": {
    "type": "Point",
    "coordinates": [
      -90.76839,
      38.76884
    ]
  }
}
```

I don't know enough here to say if they should or should not both be pointed at this location: https://getmyvax.org/api/edge/locations/54ff60f9-0c5f-4b13-920c-c245fcf6424c

But I think for the purpose of updating geolocations, either of the longitude/latitudes here are probably fine. Lemme know what you think, though.

Mr0grog: Well, that first one clearly is *very* wrong — it’s in a different town altogether and about a mile and a half down the road from the second one (which exactly matches the location with the given ID in our DB). The first one isn’t even a CVS (A more local franchise: https://www.lslmedshoppe.com/).

That also means the `_tag_provider`, `vtrcks`, `getmyvax_org`, and `cvs` concordances on the first one are wrong. I’d bet one of the `placekey` ones is correct and the other is not. :\

Also fun: It looks like vaccines.gov has two separate entries for the first place ([4d0eed17-693f-45d7-bc2a-9f0612e6deb9](https://www.vaccines.gov/provider/?id=4d0eed17-693f-45d7-bc2a-9f0612e6deb9) and [b16487fd-4da1-4a13-bf28-ffd50a529d64](https://www.vaccines.gov/provider/?id=b16487fd-4da1-4a13-bf28-ffd50a529d64)) which we’ll have to dedupe in or after merging #312.

Anyway, this just keeps me skeptical of using our IDs from their concordances list. Since we are only looking at CVS and Rite Aid for now, the actual `cvs` or `rite_aid` concordances *seem* like better indicators to key off of (although they are obviously wrong in the example you found here, too).

astonm: Cool, will just remove the univaf ID sending altogether. Fingers crossed most of the `cvs` and `rite_aid` concordances are good?

loader/test/vts.geo.test.js
```diff
@@ -0,0 +1,70 @@
+const nock = require("nock");
+const { checkAvailability } = require("../src/sources/vts/geo");
+
+const apiResponse = require("./fixtures/vts.geo.test.json");
+const nopHandler = () => {};
```
Mr0grog, line 5: `noopHandler` ?

Mr0grog (approved):

loader/test/vts.geo.test.js
astonm: Sure, if you like 👍🏽

Mr0grog: I guess I've seen `noop` for "no-op" reasonably often (don't love it), but have never seen `nop` before. 🤷

Mr0grog: (I usually try to use complete words, so tend to use a name like `doNothing` for no-op functions.)

astonm:
Merging this PR now. Once deployed, we can run the geographic coordinate import whenever we like.

PR #309, opened 2021-07-22
Create migration to add unpadded store ids for CVS locations

### LINKED ISSUE #308: Create non-zero-padded external IDs for any existing zero-padded ones
Due to legacy reasons, we have some external IDs that are zero-padded to always be a certain length. For example, CVS IDs are 5 digits, zero padded: https://github.com/usdigitalresponse/univaf/blob/bcf5aca4e9b169723843478177973efa12bebc18/loader/src/sources/cvs/smart.js#L183-L185

The arbitrary-length zero-padded IDs can make matching data from different systems complex and requires knowing specialized things. That also makes things more error-prone.

Now that we can have multiple external IDs for the same system on a single location, we should identify all our external IDs that are numeric and zero-padded (like CVS store numbers), and make non-zero-padded ones.

For example CVS #01929 has:

```js
{
  "id": "f3eda65d-230f-4e1e-b438-f432068bca5d",
  "name": "CVS #01929",
  "external_ids": [
    ["cvs", "01929"],
    ["univaf_v0", "CVS:01929"]
  ]
}
```

And we should make it:

```js
{
  "id": "f3eda65d-230f-4e1e-b438-f432068bca5d",
  "name": "CVS #01929",
  "external_ids": [
    ["cvs", "01929"],
    ["cvs", "1929"],
    ["univaf_v0", "CVS:01929"]
  ]
}
```

astonm:
This PR creates `external_id` entries for CVS stores that currently are shaped like `07000` but also may show up in other places as `7000`. This way we catch both formats.

Fixes #308

### DIFF
--- a/server/migrations/20210722181933_unpadded_ext_ids.js
+++ b/server/migrations/20210722181933_unpadded_ext_ids.js
@@ -2,7 +2,7 @@ exports.up = async function (knex) {
   const zeroPaddedRows = await knex("external_ids")
     .select("provider_location_id", "system", "value")
     .where("system", "cvs")
-    .where("value", "~", "^0\\d+$");
+    .where("value", "~", String.raw`^0\d+$`);
 
   const unpaddedRows = zeroPaddedRows.map((r) => ({
     ...r,
--- a/server/migrations/20210722181933_unpadded_ext_ids.js
+++ b/server/migrations/20210722181933_unpadded_ext_ids.js
@@ -9,7 +9,10 @@ exports.up = async function (knex) {
     value: r.value.replace(/^0+/, ""),
   }));
 
-  return knex.batchInsert("external_ids", unpaddedRows);
+  return knex("external_ids")
+    .insert(unpaddedRows)
+    .onConflict(["provider_location_id", "system", "value"])
+    .ignore();
 };
 
 exports.down = async function () {
--- a/server/migrations/20210722181933_unpadded_ext_ids.js
+++ b/server/migrations/20210722181933_unpadded_ext_ids.js
@@ -11,9 +11,9 @@ exports.up = async function (knex) {
 
   if (unpaddedRows.length) {
     return knex("external_ids")
-    .insert(unpaddedRows)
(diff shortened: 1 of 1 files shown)


server/migrations/20210722181933_unpadded_ext_ids.js
```diff
@@ -0,0 +1,17 @@
+exports.up = async function (knex) {
+  const zeroPaddedRows = await knex("external_ids")
+    .select("provider_location_id", "system", "value")
+    .where("system", "cvs")
+    .where("value", "~", "^0\\d+$");
```
Mr0grog, line 5: I was thinking this needed to be double-escaped, but I guess not!

Side note: `String.raw` is nice for this kind of stuff:

```suggestion
    .where("value", "~", String.raw`^0\d+$`);
```

Mr0grog (approved):
Argh, I drafted a review of this last week, but apparently never clicked the submit button! This is great. So sorry for the delay.

Mr0grog:
Oh wait hold up on this

server/migrations/20210722181933_unpadded_ext_ids.js
```diff
@@ -0,0 +1,17 @@
...
+    .select("provider_location_id", "system", "value")
+    .where("system", "cvs")
+    .where("value", "~", String.raw`^0\d+$`);
+
+  const unpaddedRows = zeroPaddedRows.map((r) => ({
+    ...r,
+    value: r.value.replace(/^0+/, ""),
+  }));
+
+  return knex.batchInsert("external_ids", unpaddedRows);
```
Mr0grog, line 12: Just realized this migration can fail if the database already has some existing rows like the ones we’re creating here. It looks like `.onConflict` doesn’t work with `batchInsert`, so we probably need another way to address this, either by looping and inserting or querying for unpadded values that already exist and removing them from the list before inserting.

astonm:
@Mr0grog good catch. Updated here to use `insert` instead of `batchInsert`, and to handle insertion conflicts.

Issue #135, opened 2021-05-18
Implement loader for Walgreens SMART SL API


Mr0grog:
Walgreens’s semi-official SMART Scheduling Links API is located at `https://wbaschedulinglinks.blob.core.windows.net/fhir/$bulk-publish`; we should write a loader for it. As part of this, we should further clean up the code for CVS’s SMART Scheduling Links API so as much of its implementation as possible can be shared.

Some notes worth considering here:
- ~This is *semi-official* in that it’s based off their limited-access API and written with their consent (by Josh Mandel, who is heading up the overall SMART SL standardization effort). But~ It is not directly published by Walgreens, but is their officially designated API. (Code here: https://github.com/jmandel/wba-appointment-fetch) *(Update: Walgreens confirmed today this is their official API.)*
- It’s less detailed than VaccineSpotter’s scraping data:
    - Only one, 5-day long slot per schedule, so we don’t even get day-level granularity.
    - Slots have a capacity of 0 or 5 indicating none or *some* availability. There’s no actual count of available appointments. (This is a lot like CVS’s implementation, except this one uses `5` where CVS uses `1`. But they mean the same thing.)
    - No product or dose info (VS *usually* has those for locations with availability).
- Updates are very frequent at every 5 minutes.
- It has NPIs (National Provider Identifier). These were new to me, but might be useful.

So there are some definite trade-offs vs. VaccineSpotter here. Nick M has suggested we might consider combining the two, e.g. grab product/dose info from VS and yes/no availability from this. The downside there is that we might sometimes be wrong about the product/dose, and that we can’t say anything on a day-by-day level because we don't know which individual slots that VS is showing are still valid.

I’m going to check in w/ New Jersey folks about what they think of the tradeoffs tomorrow morning and with Walgreens folks in the national sync-up tomorrow afternoon.

---

*Update 2021-05-18: Walgreens confirmed this is the official API.*

Mr0grog:
Bumping up the priority here since VaccineSpotter is planning to shut down in a week.

PR #312, opened 2021-07-31
Add data loader for CDC

### LINKED ISSUE #300: Add data loader for CDC’s Socrata Data Portal

astonm:
This PR is an approach to loading CDC data, and works with all of the states they support. For the sake of making sure we get good quality matches, it limits its current output to locations where we know the store number. Later on, we may be able to expand the number of locations loaded when we have more Vaccines.gov UUIDs in our system (perhaps from VtS data).

Fixes #300

### DIFF
--- a/loader/src/sources/cdc/api.js
+++ b/loader/src/sources/cdc/api.js
@@ -118,7 +118,7 @@ function formatStore(storeItems) {
         available_count: getAvailableCount(productList),
         available: formatAvailable(productList),
         capacity: formatCapacity(productList),
-        products: formatProducts(productList),
+        products: formatProductTypes(productList),
         // XXX should doses be here?
       },
     };
@@ -144,15 +144,28 @@ function formatAvailable(products) {
   return getAvailableCount(products) > 0 ? Available.yes : Available.no;
 }
 
-function formatProducts(products) {
-  [...new Set(products.map((p) => p.med_name))];
+function getProductType(productName) {
+  if (productName.startsWith("Pfizer")) {
+    return "pfizer";
+  }
+  if (productName.startsWith("Moderna")) {
+    return "moderna";
+  }
+  if (productName.startsWith("Janssen")) {
+    return "jj";
+  }
+  throw new Error(`Unexpected product name '${productName}'`);
+}
+
+function formatProductTypes(products) {
+  return [...new Set(products.map((p) => getProductType(p.med_name)))];
 }
 
 function formatCapacity(products) {
   return products.map((product) => {
     const availableCount = parseInt(product.supply_level, 10);
     return {
-      products: [product.med_name],
+      products: [getProductType(product.med_name)],
       date: product.quantity_last_updated,
       available: availableCount > 0 ? Available.yes : Available.no,
       available_count: availableCount,
--- a/loader/src/sources/cdc/api.js
+++ b/loader/src/sources/cdc/api.js
@@ -55,8 +55,8 @@ function formatStore(storeItems) {
       provider: "cdc",
     });
 
-    const externalId = getExternalId(base);
-    if (!externalId) {
+    const storeExternalId = getStoreExternalId(base);
+    if (!storeExternalId) {
       return null;
     }
 
@@ -100,7 +100,10 @@ function formatStore(storeItems) {
 
     result = {
       id: `cdc:${base.provider_location_guid}`,
-      external_ids: [externalId],
+      external_ids: [
+        ["vaccines_gov", base.provider_location_guid],
+        storeExternalId,
+      ],
       name: titleCase(base.loc_name),
       provider: "cdc",
 
@@ -143,7 +146,7 @@ const systemNameRe = {
   walmart: /^Walmart/i,
 };
 
-function getExternalId(store) {
+function getStoreExternalId(store) {
   if (!store.loc_store_no || store.loc_store_no == "Not applicable") {
     return null;
   }
--- a/loader/src/sources/cdc/api.js
+++ b/loader/src/sources/cdc/api.js
@@ -1,7 +1,7 @@
 const Sentry = require("@sentry/node");
 const got = require("got");
 
-const { Available, LocationType } = require("../../model");
+const { Available } = require("../../model");
 const { titleCase } = require("../../utils");
 
 function warn(message, context) {
@@ -25,7 +25,7 @@ async function* queryState(state) {
       });
       const results = JSON.parse(response.body);
 
-      for (result of results) {
+      for (const result of results) {
         yield result;
       }
 
--- a/loader/src/sources/cdc/api.js
+++ b/loader/src/sources/cdc/api.js
@@ -55,6 +55,11 @@ function formatStore(storeItems) {
       provider: "cdc",
     });
 
+    const externalId = getExternalId(base);
+    if (!externalId) {
+      return null;
+    }
+
     const addressLines = [
       base.loc_admin_street1,
       base.loc_admin_street2,
@@ -95,7 +100,7 @@ function formatStore(storeItems) {
 
     result = {
       id: `cdc:${base.provider_location_guid}`,
-      external_ids: ["vaccines_gov", base.provider_location_guid], // XXX what's the right id to send here?
+      external_ids: [externalId],
       name: titleCase(base.loc_name),
       provider: "cdc",
 
@@ -124,6 +129,30 @@ function formatStore(storeItems) {
     };
   });
   return result;
+}
+
+const systemNameRe = {
+  costco: /^Costco/i,
+  cvs: /^CVS/i,
+  kroger: /^Kroger/i,
+  publix: /^Publix/i,
+  rite_aid: /^Rite Aid/i,
+  safeway: /^SAFEWAY/i,
+  sams_club: /^Sams Club/i,
+  walgreens: /^Walgreens/i,
+  walmart: /^Walmart/i,
+};
+
+function getExternalId(store) {
+  if (!store.loc_store_no || store.loc_store_no == "Not applicable") {
+    return null;
+  }
+
+  for (system in systemNameRe) {
+    if (store.loc_name.match(systemNameRe[system])) {
+      return [system, store.loc_store_no.replace(/[^0-9]/g, "")];
+    }
+  }
 }
 
 function formatValidAt(products) {
--- a/loader/src/sources/cdc/api.js
+++ b/loader/src/sources/cdc/api.js
@@ -258,7 +258,7 @@ async function checkAvailability(handler, options) {
     const formatted = stores
       .map(formatStore)
       .filter(Boolean)
-      .map(markUnexpectedCVSs)
+      .map(markUnexpectedCvsIds)
       .forEach((item) => handler(item, { update_location: true }));
 
     results = results.concat(formatted);
@@ -269,7 +269,7 @@ async function checkAvailability(handler, options) {
 
 // 18 locations we don't have from the official CVS API that need special treatment
 // read more here: https://github.com/usdigitalresponse/univaf/pull/312#pullrequestreview-726002380
-const unexpectedCVSs = new Set([
+const unexpectedCvsIds = new Set([
   "cvs:9773",
   "cvs:9621",
   "cvs:9541",
@@ -290,10 +290,10 @@ const unexpectedCVSs = new Set([
   "cvs:7386",
 ]);
 
-function markUnexpectedCVSs(store) {
+function markUnexpectedCvsIds(store) {
   // mutates stores that match above list to hide them and add an internal note
   for (const [system, value] of store.external_ids) {
-    if (unexpectedCVSs.has(`${system}:${value}`)) {
+    if (unexpectedCvsIds.has(`${system}:${value}`)) {
       store.is_public = false;
       store.internal_notes =
         "Exists in CDC open data but not in CVS APIs; " +
--- a/loader/src/sources/cdc/api.js
+++ b/loader/src/sources/cdc/api.js
@@ -10,7 +10,7 @@ function warn(message, context) {
 }
 
 function error(message, context) {
-  console.error(`VTS Geo: ${message}`, context);
+  console.error(`CDC API: ${message}`, context);
   Sentry.captureMessage(message, Sentry.Severity.Error);
 }
 
@@ -99,7 +99,6 @@ function formatStore(storeItems) {
     });
 
     result = {
-      id: `cdc:${base.provider_location_guid}`,
       external_ids: [
         ["vaccines_gov", base.provider_location_guid],
         storeExternalId,
@@ -108,27 +107,29 @@ function formatStore(storeItems) {
       provider: "cdc",
 
       address_lines: addressLines,
-      city: base.loc_admin_city,
+      city: titleCase(base.loc_admin_city),
       state: base.loc_admin_state,
       postal_code: base.loc_admin_zip,
-      position: {
-        longitude: base.longitude,
-        latitude: base.latitude,
-      },
       info_phone: base.loc_phone,
       info_url: base.web_address,
       meta,
 
       availability: {
-        source: "cdc",
+        source: storeExternalId[0],
         checked_at: new Date().toISOString(),
         valid_at: formatValidAt(productList),
         available_count: getAvailableCount(productList),
         available: formatAvailable(productList),
-        capacity: formatCapacity(productList),
         products: formatProductTypes(productList),
       },
     };
+
+    if (base.longitude && base.latitude) {
+      result.position = {
+        longitude: base.longitude,
+        latitude: base.latitude,
+      };
+    }
   });
   return result;
 }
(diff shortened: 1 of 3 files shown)


Mr0grog:
> I'm not sure what we're going to do about trying to match these locations to existing locations in our database

I haven’t read through the code here yet, but my thoughts here were:
- Start by only using locations that have the `loc_store_no` column filled in, and match those (requires parsing the `loc_name` column to figure out which store system it’s talking about, unfortunately).
- Match up other, non store-number locations out of band and add their vaccines.gov GUIDs as external IDs manually. After that’s done, turn this on for all rows, not just those with `loc_store_no`.

loader/src/sources/cdc/api.js
```diff
@@ -0,0 +1,282 @@
...
+    const productList = storeItems.map((item) => {
+      const out = {};
+      for (const field of productFields) {
+        out[field] = item[field];
+      }
+      return out;
+    });
+
+    result = {
+      id: `cdc:${base.provider_location_guid}`,
```
Mr0grog, line 102: I think it would be good to drop this line. The other loaders only have it for historic reasons, and it would probably be best to only set it in cases where we originally read in the value from our API.

loader/src/sources/cdc/api.js
```diff
@@ -0,0 +1,282 @@
...
+      external_ids: [
+        ["vaccines_gov", base.provider_location_guid],
+        storeExternalId,
+      ],
+      name: titleCase(base.loc_name),
+      provider: "cdc",
```
Mr0grog, line 108: This should *probably* be based on `systemNameRe` instead, so we don’t flip stores from their name (e.g. `rite_aid` to `cdc` whenever this loader runs).

loader/src/sources/cdc/api.js
```diff
@@ -0,0 +1,282 @@
...
+
+      address_lines: addressLines,
+      city: base.loc_admin_city,
```
Mr0grog, line 111: Wondering if `titleCase()` would be useful here.

loader/src/sources/cdc/api.js
```diff
@@ -0,0 +1,282 @@
...
+      state: base.loc_admin_state,
+      postal_code: base.loc_admin_zip,
+      position: {
+        longitude: base.longitude,
+        latitude: base.latitude,
+      },
```
Mr0grog, line 117: It looks like there are a handful of records with no latitude/longitude. We should probably only set `position` if those are both present.

loader/src/sources/cdc/api.js
```diff
@@ -0,0 +1,282 @@
...
+  walgreens: /^Walgreens/i,
+  walmart: /^Walmart/i,
+};
+
+function getStoreExternalId(store) {
+  const m = store.loc_store_no.match(/^[A-Z]*(\d+)$/); // handle cases like RA105587 -> 105587
+  if (!m) {
+    return null;
+  }
+  const storeNumber = m[1];
```
Mr0grog, line 153: Grrrr, looking at this again now, it’s not working well in the case where these values are VTrckS PINs instead of store numbers. (I didn’t even realize that was the case for some values in this column before!)

For VTrckS, the first 3 characters are an identifier for who issued the PIN, so in the comment above, `RA1` designates that this was issued by Rite Aid (the major pharmacies get to issue their own pins, then there are ones issued by CDC [not sure on the prefix there] and by state departments of health, e.g. NJ’s is `NJ<number>`). I have no idea where any official documentation for that is; I kept trying to find it but ultimately only gleaned this info through several conversations with CDC and DoH folks.

So anyway, the Rite Aid store number we should wind up with when handed `RA105587` is actually `5587` (so we also need to remove the leading zeroes to make it as normal as possible). (See our existing record for this example: https://getmyvax.org/api/edge/locations/rite_aid:5587)

loader/src/sources/cdc/api.js
```diff
@@ -0,0 +1,282 @@
...
+  // Note: this normalization makes consistent string values, not NDCs
+  const parsed = ndcCode.match(/^(\d+)-(\d+)-(\d+)$/);
+  if (!parsed) {
+    throw new Error(`Unexpected NDC format '${ndcCode}'`);
+  }
+  return parsed
+    .slice(1)
+    .map((n) => parseInt(n, 10).toString())
+    .join("~");
+}
```
Mr0grog, line 216: Not a big deal, but it *might* be worthwhile to move this NDC stuff up to `utils` or its own module or something.

loader/src/sources/cdc/api.js
```diff
@@ -0,0 +1,282 @@
...
+      info_url: base.web_address,
+      meta,
+
+      availability: {
+        source: "cdc",
+        checked_at: new Date().toISOString(),
+        valid_at: formatValidAt(productList),
+        available_count: getAvailableCount(productList),
+        available: formatAvailable(productList),
+        capacity: formatCapacity(productList),
```
Mr0grog, line 128: I’m not really sure we should set the `capacity` field… since this isn’t really appointments, we shouldn’t try to guess something here. Better to just set the top level `available` and `available_count` fields, I think.

Mr0grog (commented):
The only real big deal here is the store number thing. I can’t believe they put VTrckS PINs in there sometimes and store numbers other times. ugh.

loader/src/sources/cdc/api.js
```diff
@@ -0,0 +1,282 @@
+const Sentry = require("@sentry/node");
+const got = require("got");
+
+const { Available } = require("../../model");
+const { titleCase } = require("../../utils");
+
+function warn(message, context) {
+  console.warn(`VTS Geo: ${message}`, context);
```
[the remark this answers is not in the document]
astonm: Oops! Thanks.

loader/src/sources/cdc/api.js
astonm: Yeah, checking the data set, there are plenty of city names that have wonky formatting, thanks.

loader/src/sources/cdc/api.js
astonm: I have some hesitation here just because I think this code is a bit hacky and not so robust to arbitrary NDC formats.

Mr0grog: That’s fair, and a good reason to keep it here for now.

loader/src/sources/cdc/api.js
astonm: Just took a stab at handling things both ways, using the guidelines you shared in Slack

> The format so far as I understand it (may not be 100% correct):
> - Always (?) 8 characters.
> - First 3 characters designates who issued the PIN (major pharmacies, states, CDC directly, not sure if there are other issuers/assigners/whatever)
> - States are <2-letter postal abbreviation><digit>
> - Pharmacies are <2-letter abbreviation>1 or <3-letter abbreviation>. The remaining characters are usually the store number, but are really up to the pharmacy as the issuer.
> - Remaining characters are up to the issuer and could be anything.

Mr0grog (commented):
Ugh, sorry this just keeps going. A few minor notes inline and then…

I went ahead and checked to see how many locations this creates that we don’t already have and gut-check whether that seemed right:

- `walgreens`: 186

    This actually seem fine — I sampled a bunch and they were all either locations I didn’t correctly grab from our data (because they had funky provider names like `community_a_walgreens_pharmacy`) or are honestly missing from our data (they are listed in the Walgreens SMART SL API #135). The funky-named ones also have `walgreens:` external IDs, so they should match up fine in practice.

- `cvs`: 18

    Every single one of these seems to have no stock, and since we *should* already have a comprehensive listing of locations actually administering vaccines from the various CVS APIs, I’m thinking these might be places that have never been stocked and never administered. It sucks to hardcode things like this, but maybe we should include these properties on the 18 locations:

    ```js
    {
      is_private: true,
      internal_notes: 'Exists in CDC open data but not in CVS APIs; this location is probably not actually administering vaccines.'
    }
    ```

    For reference, the 18 locations were:

    ```
    cvs:9773
    cvs:9621
    cvs:9541
    cvs:3428
    cvs:5627
    cvs:5646
    cvs:4592
    cvs:8699
    cvs:8194
    cvs:8210
    cvs:8214
    cvs:8916
    cvs:739
    cvs:6006
    cvs:2972
    cvs:558
    cvs:1883
    cvs:7386
    ```

- `safeway`: 5

    These are all to do with the messiness of brand names. :(

    `safeway:3125` is `pak_n_save` in our DB (location ID: `7bd930d2-2bce-4b47-bcf9-366c35797b13`). It turns out that Pak-N-Sav is a sub-brand of Safeway, so we should probably “fix” this a) by giving it a `safeway:` external ID manually, and b) by [special-casing `pak_n_save` in the Vaccinespotter loader](https://github.com/usdigitalresponse/univaf/blob/16ae8b4153c0a22a5565e0503b15070a07f56ee2/loader/src/sources/vaccinespotter/index.js#L323-L341) so that it also creates `safeway:` external IDs. **(That should be a separate PR, and that OR the manual fix should be done before merging this PR.)**

    The rest are listed as Alberstons stores in our DB. Checking out other sources, it looks like these were probably recently rebranded (street view photos from 2017 show them as Albertsons-branded, but one with a very recent streetview shows Safeway, and the store listings on Albertsons’s website shows them as Safeways. We should probably just manually add `safeway:` external IDs for these in the database.

    - `safeway:414` is in our DB as `daeed1fd-ed16-4438-9a6c-b794f343a567`
    - `safeway:2248` is in our DB as `b70c960b-c6e5-4043-a1cd-c1a4e9cc5b28`
    - `safeway:469` is in our DB as `7d196a9e-c781-4fed-9705-22cebb3ec645`
    - `safeway:186` is in our DB as `a4ba3ad1-f746-4945-983e-99b8100c26de`

- `publix`: 15

    These all appear to be legitimately missing from our DB.

Mr0grog:
The reverse (locations we’ve got but the CDC data set doesn’t) also shows some disparity we should investigate, but I don’t think is a blocker for this PR:

```js
{
  cvs: 24,
  walgreens: 295,
  kroger: 15,
  publix: 26,
  costco: 2,
  safeway: 5,
  rite_aid: 2
}
```

Mr0grog:
Also, just realized I was probably not super clear last night on what I think needs to happen w/r/t the locations that looked problematic to me, so:

- Nothing to do for the `walgreens` and `publix` ones. The disparity looks correct, so this loader will do the right thing.
- The 5 `safeway` locations:
    - Need manual remediation in our DB before deploying this.
    - For the first one, the Vaccinespotter loader should also be amended to automatically set the right external ID and location name and provider, but as a separate PR from this.
    - Not sure there’s anything we can do in the Vaccinespotter loader for the other 4.
- The 18 CVS locations should probably be hard-coded to not be public. I’m not 100% on that, though, and am curious what you think.

loader/src/sources/cdc/api.js
```diff
@@ -0,0 +1,270 @@
...
+  console.warn(`CDC API: ${message}`, context);
+  Sentry.captureMessage(message, Sentry.Severity.Info);
+}
+
+function error(message, context) {
+  console.error(`CDC API: ${message}`, context);
+  Sentry.captureMessage(message, Sentry.Severity.Error);
+}
+
+async function* queryState(state) {
```
jaronoff97, line 17: Never seen this generator syntax before, very cool!!

astonm:
@Mr0grog thanks for all the feedback here. I trust your instincts on the CVS situation–seems like we can safely hide them for now.

loader/src/sources/cdc/api.js
```diff
@@ -0,0 +1,308 @@
...
+
+    results = results.concat(formatted);
+  }
+
+  return results;
+}
+
+// 18 locations we don't have from the official CVS API that need special treatment
+// read more here: https://github.com/usdigitalresponse/univaf/pull/312#pullrequestreview-726002380
+const unexpectedCVSs = new Set([
```
Mr0grog, line 272: Very minor nit: everywhere else, we treat acronyms as words in camel case, so this should be `unexpectedCvses` or maybe `unexpectedCvsIds`.

loader/src/sources/cdc/api.js
```diff
@@ -0,0 +1,308 @@
...
+  [normalizeNDC("59676-0580-05")]: "jj", //use
+  [normalizeNDC("59676-0580-15")]: "jj", //sale
+  [normalizeNDC("80631-0100-01")]: "novavax", //sale
+  [normalizeNDC("80631-0100-10")]: "novavax", //use
+  [normalizeNDC("80777-0273-10")]: "moderna", // use
+  [normalizeNDC("80777-0273-15")]: "moderna", // use
+  [normalizeNDC("80777-0273-98")]: "moderna", //sale
+  [normalizeNDC("80777-0273-99")]: "moderna", //sale
+};
+function normalizeNDC(ndcCode) {
```
Mr0grog, line 213: Saw the CVS one and realized I totally missed the same capitalization issue here.

loader/src/sources/cdc/api.js
```diff
@@ -0,0 +1,308 @@
...
+]);
+
+function markUnexpectedCVSs(store) {
+  // mutates stores that match above list to hide them and add an internal note
+  for (const [system, value] of store.external_ids) {
+    if (unexpectedCVSs.has(`${system}:${value}`)) {
+      store.is_public = false;
+      store.internal_notes =
+        "Exists in CDC open data but not in CVS APIs; " +
+        "this location is probably not actually administering vaccines.";
```
Mr0grog, line 300: The `utils.oneLine` tag function is great for this. 🙂

Mr0grog (commented):
Just style issues.

loader/src/sources/cdc/api.js
```diff
@@ -0,0 +1,309 @@
...
+  for (const product of products) {
+    const supplyLevel = parseInt(product.supply_level, 10);
+    if (supplyLevel > 0) {
+      sum += supplyLevel;
+    } else {
+      sum += product.in_stock ? 1 : 0;
+    }
+  }
+  return sum;
+}
```
Mr0grog, line 190: Welp, I discovered some documentation we must have missed today that explains this, and the `supply_level` column is not an actual count in any way. It’s an enum describing *how long current supply should last*. https://data.cdc.gov/api/views/5jp2-pgaw/files/2a8f085d-04ab-4044-8148-3b7df4cb7088?download=true&filename=Outbound%20File%20Specification.pdf

| Value | Definition |
| ----: | ---------- |
| -1 | No report |
| 0 | No supply |
| 1 | < 24 hours of supply |
| 2 | 24–48 hours of supply |
| 3 | > 24 hours of supply |

…so we should probably only be using this to infer yes/no for `available`, and not try to set `available_count` with it at all.

astonm: 😮 

Will fix, thanks.

astonm:
@Mr0grog I think this is already to after we update the database. Here's the SQL I have planned:

```
INSERT INTO external_ids (provider_location_id, system, value) VALUES 
 ('7bd930d2-2bce-4b47-bcf9-366c35797b13', 'safeway', '3125'),
 ('daeed1fd-ed16-4438-9a6c-b794f343a567', 'safeway', '414'),
 ('b70c960b-c6e5-4043-a1cd-c1a4e9cc5b28', 'safeway', '2248'),
 ('7d196a9e-c781-4fed-9705-22cebb3ec645', 'safeway', '469'),
 ('a4ba3ad1-f746-4945-983e-99b8100c26de', 'safeway', '186');
```

I'm thinking we just run it manually, but I can add a migration if you prefer.

Mr0grog:
That looks good to me. I think running it manually makes sense and is a good way to go. 👍

Mr0grog (approved):
Looks good to me! Might be nice to have some simple tests, but this is probably also not worth holding up for that (we can do them in a second pass as we work to support for location types).

Feel free to merge after to running the SQL. 👍

Mr0grog:
Oh, and I guess we'll need a quick follow-on to schedule this in Terraform. (1/day?)

PR #314, opened 2021-07-31
Support SMART Scheduling Links API for PrepMod

### LINKED ISSUE #313: Implement SMART loader for PrepMod
PrepMod now supports the SMART Scheduling Links API. It includes product, but not dose information. The API is open-access.

Each PrepMod install is a separate application, and the API is at:

```
<prepmod-host>/api/smart-scheduling-links/$bulk-publish
```

For example, the State of Washington’s PrepMod API is at:

```
https://prepmod.doh.wa.gov/api/smart-scheduling-links/$bulk-publish
```

Mr0grog:
This is a first pass and rough implementation of SMART Scheduling Links for PrepMod. It has a lot of code duplicated from the CVS SMART loader, and some other rough spots, but does work reasonably well.

Fixes #313.

To Do:

- [x] This needs at least some tests.
- [ ] Deduplicate repeated code from CVS (on the flip side, there are some other SMART loaders still to be implemented, and it might be better to wait for a third implementation before trying to smooth over the differences).


loader/src/sources/prepmod/hosts.js
```diff
@@ -0,0 +1,77 @@
...
+    _: "https://hampton-va.cdn.prod.prepmodapp.com",
+  },
+
+  // TX university
+  "Prairie View A&M University": {
+    _: "https://pvam.cdn.prod.prepmodapp.com",
+  },
+};
+
+module.exports = prepmodHostsByState;
```
astonm, line 77: nit: It might be better to export `{ prepmodHostsByState }` instead to give us flexibility if we ever want to export something else alongside.

loader/src/sources/prepmod/index.js
```diff
@@ -0,0 +1,427 @@
...
+  210: "astra_zeneca",
+  207: "moderna",
+  211: "novavax",
+  208: "pfizer",
+  212: "jj",
+};
+
+// Use symbols to link slots -> schedules -> locations to avoid circular
+// references when serializing or logging data.
+const locationReference = Symbol("schedule");
```
astonm, line 39: Doesn't affect functionality, but would increase clarity:

```suggestion
const locationReference = Symbol("location");
```

loader/src/sources/prepmod/index.js
```diff
@@ -0,0 +1,427 @@
...
+    .split("\n")
+    .filter(Boolean)
+    .map((line, index) => {
+      try {
+        return JSON.parse(line);
+      } catch (error) {
+        throw new SyntaxError(`Error parsing line ${index}: ${line}`);
+      }
+    });
+}
```
astonm, line 58: Probably worth adding to `src/utils.js`?

loader/src/sources/prepmod/index.js
```diff
@@ -0,0 +1,427 @@
...
+    const manifest = await this.getManifest();
+    for (const item of manifest.output) {
+      if (item.type === "Location") {
+        const response = await got(item.url);
+        for (const location of parseJsonLines(response.body)) {
+          yield location;
+        }
+      }
+    }
+  }
```
astonm, line 94: Maybe me trying to abstract too much here, but it seems like it might be useful to have an `async *listItems(itemType)` that applies a simple `item.type` filter that would turn `listLocations`, `listSchedules` and `listSlots` into one-liners.

Although now that I say it, why not pull all three at the same time?

astonm (approved):
Looks pretty good to me. Just a few ideas below for you about small refactors, if you like.

loader/src/sources/prepmod/index.js
Mr0grog: Gah! This must have been a copy/paste error.

loader/src/sources/prepmod/index.js
Mr0grog: Ah, good idea on abstracting the filtering logic to an underlying method! I didn’t pull them all at once because we need to parse all of one type before parsing the next types. (But `getManifest()` has caching, so the separate method calls still only make one HTTP request.)

Mr0grog:
Followed up on all the comments. Not going to try and de-dupe with the other SMART SL code. Merging if this passes all checks.

PR #321, opened 2021-08-03
[305] Add Winston Logging Library for Server

### LINKED ISSUE #305: Use a real logging library in the server
*I was totally certain I’d written an issue for this in Monday’s meeting, but it turns out that was a lie. The only mention I can find is a side-note in https://github.com/usdigitalresponse/univaf/issues/52#issuecomment-820827264) and [a to-do comment in the code](https://github.com/usdigitalresponse/univaf/blob/bcf5aca4e9b169723843478177973efa12bebc18/server/src/app.ts#L15-L16).*

We currently use `console.log` to do any logging we want in the API server, but we should really use a logging library that lets us set logging levels and manage output more carefully. While we’re at it, we should set [per-request logs](https://github.com/usdigitalresponse/univaf/blob/bcf5aca4e9b169723843478177973efa12bebc18/server/src/app.ts#L15-L29) at a level that does not log in production, but does log in dev environments, since our production use cases are better handled with metrics and DataDog right now.

I think [Winston](https://www.npmjs.com/package/winston) is a good, reliable, mature choice for this, but there are definitely plenty of other options here, since Node.js really has nothing built-in. (Winston is what I helped NJ implement for theirs.)

jaronoff97:
This PR adds in the Winston logging library rather than using the default console. We are still outputing to console by default for now, but Winston gives us a lot of flexibility for the future. The default log level is override, but one can set "LOG_LEVEL=debug" to enable debug logs now. The only functional change here is that we will no longer log every request we see.

This PR will close #305

## Example with debug enabled

Screen Shot 2021-08-03 at 4 01 14 PM

## Example without log level override

Screen Shot 2021-08-03 at 4 01 30 PM


server/src/app.ts
```diff
@@ -128,7 +120,7 @@ smartSchedulingApi.use((_req: Request, res: Response) =>
 smartSchedulingApi.use(Sentry.Handlers.errorHandler());
 smartSchedulingApi.use(
   (error: any, req: Request, res: Response, _next: NextFunction) => {
-    console.error("ERRROR:", error);
+    logger.error("ERRROR:", error);
```
Mr0grog, line 123: I’m not sure it really makes sense to include the `ERROR` text here since the format you set up already prefixes this with the level. It doesn’t really seem useful to log `error: ERROR: {error}`.

This is also not really correct — `console.*()` takes N arguments that are formatted as appropriate for their type (e.g. strings are text, but objects get all their properties printed out), so this made sense when it was `console.error`. But Winston loggers have different arguments: `logger.*(message, objectWithMoreInfoForFormatting)`. It so happens that this *sort* of works because an error’s text is in the `message` property, and Winston will concatenate any `message` property in the info object with the message argument. But I don’t think you meant to do that (and if you did, I think we should be more explicit about it; this code is leveraging some serious subtleties).

server/src/app.ts
```diff
@@ -147,7 +139,7 @@ if (app.get("env") === "development") {
   app.use(errorHandler());
 } else {
   app.use((error: any, req: Request, res: Response, _next: NextFunction) => {
-    console.error("ERROR:", error);
+    logger.error("ERROR:", error);
```
Mr0grog, line 142: See above about redundant level info and different arguments between `console.error` and `logger.error`.

server/src/server.ts
```diff
@@ -20,12 +21,13 @@ process.on("SIGINT", () => {
  * Start Express server.
  */
 const server = app.listen(app.get("port"), () => {
-  console.log(
-    "  App is running at http://localhost:%d in %s mode",
+  logger.log(
+    "info",
+    "App is running at http://localhost:%s in %s mode",
     app.get("port"),
     app.get("env")
```
Mr0grog, line 28: Not a blocker, but I think we’d be better off to modernize this line by using a template string instead of interpolation, which is very sensitive to argument order, slower, and requires adding the `splat()` add-on to the logger.

```suggestion
  logger.info(`App is running at http://localhost:${app.get("port")} in ${app.get("env")} mode`);
```

Mr0grog (changes requested):

server/src/logger.ts
```diff
@@ -0,0 +1,23 @@
+import { createLogger, Logger, transports, format } from "winston";
+const { combine, timestamp, splat, printf, label } = format;
+
+const level = process.env.LOG_LEVEL || "info";
```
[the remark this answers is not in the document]
Mr0grog: Oh, this is not what I meant. I was just referring to this highlighted line — we should avoid reading environment variables and external configuration sources outside the `config` module wherever possible. It was good that you had a module dedicated to the logger.

server/test/config.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  next();
+};
+const streamTransport = new transports.Stream({ stream });
+logger.add(streamTransport);
+
+describe("logger functionality", () => {
+  afterEach(() => {
+    output = "";
+  });
+  it("should print info correctly", () => {
```
Mr0grog, line 18: Minor nit: please include a blank line between each hook and each test.

server/src/config.ts
```diff
@@ -1,5 +1,22 @@
 import { Request } from "express";
 import type { Knex } from "knex";
+import { createLogger, Logger, transports, format } from "winston";
+const { combine, timestamp, splat, printf, label } = format;
+
+const level = process.env.LOG_LEVEL || "info";
```
Mr0grog, line 6: Ahhh, I was just trying to say reading external configuration data (i.e. determining the current logging level from `process.env.LOG_LEVEL`) should be in the `config` module, not the entirety of all the logging code!

jaronoff97: ah no worries! I can split it back up

server/test/logger.test.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Writable } from "stream";
+
+let output = "";
+const stream = new Writable();
+stream._write = (chunk, encoding, next) => {
+  output = output += chunk.toString();
+  next();
+};
+const streamTransport = new transports.Stream({ stream });
+logger.add(streamTransport);
```
Mr0grog, line 12: Quick last minute thought — you could clean up the test output by removing the current transports:

```suggestion
for (const transport of logger.transports.slice()) {
  logger.remove(transport);
}
logger.add(streamTransport);
```

Mr0grog (approved):
Looks good! Dropped one minor thought inline that could make test output cleaner, but not a deal breaker, and if you want to add that, no need to re-review before merging.

PR #328, opened 2021-08-09
Upgrade Sentry client in server to v6.11.0

Mr0grog:
Supersedes #318, #319.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,8 +60,6 @@ jobs:
 
       - name: Run server tests
         run: |
-          printenv
-          echo '-------------------------'
           cd server
           npm run test
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,8 @@ jobs:
 
       - name: Run server tests
         run: |
+          printenv
+          echo '-------------------------'
           cd server
           npm run test
         env:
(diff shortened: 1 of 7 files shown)


Mr0grog:
Welp, it seems like something in this version of sentry is causing tests to hang in CI, but not locally. :\

Mr0grog:
Alright, I understand this pretty clearly now; it’s a regression caused by `autoSessionTracking` being turned on by default in Sentry 6.9+. The session tracking infrastructure runs forever until you explicitly shut down Sentry, and therefore keeps the tests open.

The best solution here is probably to explicitly set `autoSessionTracking: false`, since that reverts the same behavior we had before v6.9.

astonm:
Oof. How would you feel about skipping Sentry initialization in tests?

Mr0grog:
I’m actually not sure what would happen if we didn’t call `Sentry.init()` at all. 🤔

But since `autoSessionTracking` used to default to `false`, I think just explicitly setting it to `false` now is probably the best solution. It makes for less fuss in the code, at least.

Mr0grog:
As a side note, can you repro the hang locally (using the first commit of this PR), @astonm? I could not, even though looking at the Sentry code I can’t see what would be happening differently. Maybe some environment variable in CI.

astonm:
@Mr0grog I can't get the hang in tests to repro locally either.

Mr0grog:
Nevermind, I see what’s happening now — if you don’t specify a `release` option, Sentry automatically turns off `autoSessionTracking`. But it detects the value from `release` from a huge variety of possible env vars, including… `GITHUB_SHA`, which is set in the GitHub Actions environment.

(And experimentally, if I set that locally, tests hang. So Sentry was implicitly turning the feature *on* in test and *off* in my local environment. Ugh.)

PR #339, opened 2021-08-18
Use correctly formatted `position` in `/api/edge/locations/<external_id>`

Mr0grog:
When using an external ID (as opposed to an actual UNIVAF ID) in the `/api/edge/locations/<id>` endpoint, we are currently returning the `position` value as WKB:

```js
// http://getmyvax.org/api/edge/locations/rite_aid:5610
{
  "data": {
    "id": "64d41035-6dcf-4331-a9f9-438dec46ee31",
    "position": "0101000020E6100000D122DBF97E7A5DC08716D9CEF7034140",
    // ...
  }
}
```

It looks like we introduced this bug in #147, and I noticed it while reviewing #304. This fixes it and adds coverage for it in tests.

### DIFF
--- a/server/src/db.ts
+++ b/server/src/db.ts
@@ -397,7 +397,9 @@ export async function getLocationByExternalIds(
       fields.map((name) => {
         name = `provider_locations.${name}`;
         // Ensure we return geo coordinates in an easy-to-handle format.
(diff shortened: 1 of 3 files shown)


astonm (approved):
Oops. lgtm.

PR #349, opened 2021-09-01
Load PrepMod data from Washington and Alaska

Mr0grog:
Now that the PrepMod loader works, let's start loading it! For now, the API is only turned on in Washington, and we are actively working to get it turned *back* on in Alaska, so this just runs it in those two states.


astonm (approved):

Mr0grog:
Thanks!

PR #356, opened 2021-09-15
Handle invalid `address_lines` data (server and loader)

### LINKED ISSUE #355: Vaccine Spotter loader sends invalid `address_lines` field
The API received buggy data from the `vaccinespotter` loader that looks like the following:

```json
{
  "address_lines": "",
  ...
}
```

where `address_lines` is expected to be an array instead. We should make sure to send a valid value here (e.g. `[]`), but also may want to beef up validation on the server to ensure that the error comes from higher up the stack than the database library.

*Issue found via Sentry: https://sentry.io/organizations/usdr/issues/2602455976/?project=5718480*

astonm:
This PR ensures that falsy values for `address_lines` don't cause unexpected errors by adding some server-side validation to coalesce those values into an empty array. It appears the only loader sending that sort of bad data was Vaccine Spotter, so we also apply a small fix there to clean up its output.

Fixes #355 from the server side as well as the loader side. Either fix would have sufficed, but both were easy, so 🤷🏽‍♂️.

### DIFF
--- a/loader/src/sources/vaccinespotter/index.js
+++ b/loader/src/sources/vaccinespotter/index.js
@@ -256,14 +256,17 @@ const formatters = {
       id = `vaccinespotter:${store.properties.id}`;
     }
 
+    let addressLines = [];
+    if (store.properties.address) {
+      addressLines.push(titleCase(store.properties.address));
+    }
+
     return {
       id,
       location_type: LocationType.pharmacy,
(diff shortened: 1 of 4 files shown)


Mr0grog (changes requested):
Nice! A couple minor issues inline.

Mr0grog (approved):
👍 :shipit:

PR #357, opened 2021-09-17
Only specify a major version for the database engine

astonm:
SSIA.


Mr0grog (approved):
This looks good to me! We should keep a close eye on this when we apply to make sure it doesn’t actually change the RDS instance (I’m a little disappointed that in the TF plan it shows as a change to the DB, but 🤷 oh well).

Mr0grog:
Per discussion on Slack, I’m going ahead and merging this. Will keep an eye on what actually happens.

Mr0grog:
Applied, and seems to have worked as expected! The DB does not appear to have gone down for any period of time, and we are still on PG v13.3.

PR #370, opened 2021-09-23
Only show in-stock products from CDC loader

### LINKED ISSUE #369: CDC loader is always listing all products as available
@dhcole in New Jersey noted that the CDC loader appears to be listing all product types that a location might ever supply in the `availability.products` field, even if those products are contributing to the value of `availability.available`. i.e. we are listing every product as available, regardless of current stock levels.

I don’t remember if we did that consciously (out of concern that supply is a deeply inaccurate proxy for appointment availability) or not, but being consistent with the way we determine the overall YES/NO for `available` is probably a better approach, and what he would have expected.

Example case: only Pfizer is in stock at this location right now… https://www.vaccines.gov/provider/?id=d313cdc7-0ddb-4c38-8bb5-31dc71a5c95a&inventoryCategory=covid&zipcode=07011&medications=779bfe52-0dd8-4023-a183-457eb100fccc%2Ca84fb9ed-deb4-461c-b785-e17c782ef88b%2C784db609-dc1f-45a5-bad6-8db02e79d44f&radius=25&appointments=true

…but our API lists all three: https://getmyvax.org/api/edge/locations/f3eda65d-230f-4e1e-b438-f432068bca5d

Mr0grog:
This changes the CDC loader to only list products that are in stock. Fixes #369.

This adds a single test case to cover this particular situation, though it would be good to do another pass in the future to cover things more expansively. It also makes another minor fix where the return value of `checkAvailability()` was incorrect (it didn’t matter that much in production because callbacks were still correct, and that’s what usually is relied on).


loader/src/sources/cdc/api.js
```diff
@@ -243,8 +254,9 @@ async function checkAvailability(handler, options) {
     const formatted = stores
       .map(formatStore)
       .filter(Boolean)
-      .map(markUnexpectedCvsIds)
-      .forEach((item) => handler(item, { update_location: true }));
+      .map(markUnexpectedCvsIds);
+
+    formatted.forEach((item) => handler(item, { update_location: true }));
```
astonm, line 259: Oops! Thanks for the fix here.

astonm (approved):
lgtm.

Appreciate the cleanup you did along the way, as well.

Issue #358, opened 2021-09-17
Audit Express for automatic/debug routes that might expose sensitive data


Mr0grog:
Some web frameworks come with built-in debug routes or error handlers that could expose sensitive data, like environment variables or application globals. I don’t *think* this is an issue with express (at least when `NODE_ENV=production`), but we should double-check to make sure.

Mr0grog:
Dug into this and Express does not add any routes, but does include stack traces an sensitive info in error handles unless in `production` mode. We *are* correctly putting the app in `production` mode, so we are good there.

We use the `errorHandler` middleware which shows sensitive data, but only in `development` mode. Otherwise, we use our own handler (which I have also checked and is doing the right thing).

Two other things that I noticed here, though:
1. Express security docs recommend disabling the `X-Powered-By` header (https://expressjs.com/en/advanced/best-practice-security.html#at-a-minimum-disable-x-powered-by-header). That won’t stop someone specifically interested in our app (since they can browse our source code), but it matters in preventing drive-by attacks from people doing IP & port scanning across the web.
2. We have a `/debugme` endpoint that throws an error. I think I added this to test our error handlers, but it probably shouldn’t be enabled in production.

Will file PRs for the above items.

Mr0grog:
See #375 and #376 for the above issues.

PR #375, opened 2021-09-27
Disable `X-Powered-By` HTTP Header

Mr0grog:
Don't advertise that we are using Express, in case someone is scanning for known vulnerable servers and there's a 0-day or other known-but-not-yet-patched vulnerability.

Fixes the first part of #358.


astonm (approved):
🚢

PR #376, opened 2021-09-27
Disable `/debugme` route in production

Mr0grog:
This route is meant for testing error handlers, and should not actually be available in production. (It doesn’t expose anything sensitive, but it can spam Sentry and make us thing something’s broken when it’s not.)

Fixes the second part of #358, alongside #375.


astonm (approved):
👍🏽

PR #388, opened 2021-10-04
HOTFIX: Bump max request body size in API server

### LINKED ISSUE #387: Data from PrepMod loader is fequently too big
The PrepMod loader is currently encountering errors because the request body it is sending to the API (to get saved in the DB) is too big! Example error:

```
Error sending: request entity too large {"status_code":413,"source":"univaf-prepmod","location_name":"Discovery Health MD, Puyallup COVID Vaccine Site"}
```

The current request body size limit is 500 kB: https://github.com/usdigitalresponse/univaf/blob/21a32c6af3ae4335a703ec186dee0db872605246/server/src/app.ts#L51

We can probably try increasing that a bit, but it’s also pretty big — all that data makes SQL joins in our API slower and makes responses in requests to our own API big and slow. We may want to consider somehow abbreviating this data. :\

Mr0grog:
A few locations from the PrepMod loader are too big for our current 500 kB limit, so this is the quickest and straightest solution. We might want to do a second pass on the loader where we summarize the data more briefly, since this is a lot to ship *out* and to do joins on in our own API.

Fixes #387.

### DIFF
--- a/server/src/app.ts
+++ b/server/src/app.ts
@@ -48,7 +48,7 @@ app.disable("x-powered-by");
 app.use(Sentry.Handlers.requestHandler());
 app.use(logRequest);
 app.use(compression());
-app.use(bodyParser.json({ limit: "500kb" }));
+app.use(bodyParser.json({ limit: "1mb" }));
 app.use(cors());
 app.use(authorizeRequest);
 app.use(datadogMiddleware);


Mr0grog:
Taking thumbs-up on Slack as approval to merge here.

PR #432, opened 2021-10-29
Always update a location's `external_ids`

Mr0grog:
Even if we're not updating the other properties of a location (name, address, etc.), we should always add any new external IDs we see for a location. This is a more reasonable thing than it used to be, since you can now only add to the list for a given location, and not change existing entries.

As we add and clean up a bunch of sources from VaccineSpotter, I’m a little worried we aren’t always capturing all the new IDs we’ve seen, and aren’t able to de-duplicate as effectively as we might otherwise (if one source silently fails to add a new ID, and another source uses it, we could wind up creating a new location record). A bunch of sources already set `update_location=true` ([for example, CDC](https://github.com/usdigitalresponse/univaf/blob/6eade8ea037b32bbecea074fe9d1d0eb2387859f/loader/src/sources/cdc/api.js#L332)), so this won’t impact all sources.

If something goes horribly awry, we can roll back this change and revert to the most recent DB snapshot. I don’t think it’s too likely we’ll need that, though.

### DIFF
--- a/server/src/api/edge.ts
+++ b/server/src/api/edge.ts
@@ -264,6 +264,10 @@ export const update = async (req: AppRequest, res: Response): Promise<any> => {
       await db.updateLocation(location, data);
       result.location.action = "updated";
     }
+  } else if (data.external_ids?.length) {
+    // Otherwise just update the external IDs. This ensures we track every
+    // relevant external ID we've seen for a location.
+    await db.addExternalIds(location.id, data.external_ids);
   }
 
   if (data.availability) {
--- a/server/test/api.edge.test.ts
+++ b/server/test/api.edge.test.ts
@@ -481,6 +481,25 @@ describe("POST /api/edge/update", () => {
     ]);
   });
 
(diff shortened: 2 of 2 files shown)


Mr0grog:
Example of why this is needed: we used to get Albertsons from VaccineSpotter and only stored the 8-digit Albertsons-family-wide ID, but CDC data for Albertsons has the actual Abertsons-brand store number.

The new Albertons source gets the brand-centric store number: https://github.com/usdigitalresponse/univaf/blob/5f4db7720b1ca8fcdc3c2db2953a9ed50a582064/loader/src/sources/albertsons/index.js#L268-L281

But it doesn’t get saved if we know about the location already, and we know about most locations. The Albertsons source does not set `update_location=true` when sending updates because our old store names are often more accurate. 😬

So we’re able to extract the data that would let us match these records, but we aren’t currently able to find a good way to save it.

astonm (approved):
👍🏽

PR #435, opened 2021-10-30
Work in progress on a source for H-E-B

### LINKED ISSUE #421: Port H-E-B Loader From VaccineSpotter
H-E-B has a lot of locations that we would really like to keep working. The loader is fairly straightforward, so we should port it from VaccineSpotter: https://github.com/GUI/covid-vaccine-spotter/tree/main/src/providers/Heb

See also #418.

This is somewhat speculative! If it proves too hard to get working in practice, we can drop it.

Mr0grog:
This is based on the code from VaccineSpotter. It’s a very incomplete work-in-progress; I need to step away for the day and am pushing this up in case I don’t get to it by Monday, at which point I may not be able to do any more.

Fixes #421.

### DIFF
--- a/loader/package-lock.json
+++ b/loader/package-lock.json
@@ -19,6 +19,9 @@
         "yargs": "^17.2.1"
       },
       "devDependencies": {
+        "ajv": "^8.6.3",
+        "ajv-formats": "^2.1.1",
+        "ajv-keywords": "^5.0.0",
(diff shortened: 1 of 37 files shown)


cmroberts:
Well, I broke linting and also munged the rebase here; closing out this PR and I'll clean up this branch on Monday along with adding a nock test and validAt field before submitting a new PR

PR #436, opened 2021-10-30
Add a source for HyVee

### LINKED ISSUE #422: Port HyVee Loader From VaccineSpotter
HyVee has a lot of locations that we would really like to keep working. The loader does not appear too complex, so we should port it from VaccineSpotter: https://github.com/GUI/covid-vaccine-spotter/tree/main/src/providers/HyVee

See also #418.

This is somewhat speculative! If it proves too hard to get working in practice, we can drop it.

Mr0grog:
This is based on VaccineSpotter's code. It’s a very incomplete work-in-progress; I need to step away for the day and am pushing this up in case I don’t get to it by Monday, at which point I may not be able to do any more.

Fixes #422.


loader/src/sources/hyvee/index.js
```diff
@@ -0,0 +1,204 @@
...
+const PROVIDER_NAME = "hyvee";
+
+function warn(message, context) {
+  console.warn(`HyVee: ${message}`, context);
+  // Sentry does better fingerprinting with an actual exception object.
+  if (message instanceof Error) {
+    Sentry.captureException(message, { level: Sentry.Severity.Info });
+  } else {
+    Sentry.captureMessage(message, Sentry.Severity.Info);
+  }
```
astonm, line 25: This would make for some good utility code in a later PR.

astonm (approved):
👍🏽

astonm:
P.S. great branch name!

PR #437, opened 2021-10-30
Add schemas for use in tests

Mr0grog:
This adds an `expect(x).toMatchSchema(someJsonSchemaDefinition)` matcher to Jest in the loader project, and should be helpful in writing better tests.

There is a very nice [`jest-json-schema` package on NPM](https://www.npmjs.com/package/jest-json-schema), but it is [stuck on a several-major-releases-old version of AJV](https://github.com/americanexpress/jest-json-schema/pull/26). I wanted our schemas in the loader and server to be similar, and not have some weird edge cases because of AJV versioning problems, so I wound up just implementing something custom. At least it’s not super complicated, but still makes me sad. :(

The schemas are not the same as the server because:
- They needed to be more strict. The server ones are designed to help massage acceptable data into correct data, and here, we want to be really narrow about what we do.
- The server schemas *manipulate* the data (again, aiding in cleaning up and reshaping roughly correct data), which is not what we want here.

Ideally, the server would somehow build on/extend/inherit from these more strict schemas, but that felt like a lot more work to do and other things (https://github.com/usdigitalresponse/univaf/milestone/5) are much higher priority at the moment.

**TO-DO:** should probably actually use these!

### DIFF
--- a/loader/package-lock.json
+++ b/loader/package-lock.json
@@ -19,6 +19,9 @@
         "yargs": "^17.2.1"
       },
       "devDependencies": {
+        "ajv": "^8.6.3",
+        "ajv-formats": "^2.1.1",
+        "ajv-keywords": "^5.0.0",
         "eslint": "^7.32.0",
         "jest": "^27.2.4",
         "nock": "^13.1.0",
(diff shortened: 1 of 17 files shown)


astonm (approved):
Cool, looks pretty simple but flexible!

Mr0grog:
Implemented these in some tests and discovered a few subtle bugs and some things that weren’t quite right about the schema!

10xLaCroixDrinker:
Hey @Mr0grog, we updated jest-json-schema to use ajv@8 in v6.0.0

PR #438, opened 2021-10-30
Fix Nock-Back issues discovered with schema test

Mr0grog:
**This depends on #437.**

I discovered some issues with our Nock-Back plugin when I tried to *use* #437, and tried to find some solutions here. This might be OK, or might need a *little* more tidying up before merging.

### DIFF
--- a/loader/test/support/nock-back-plugin.js
+++ b/loader/test/support/nock-back-plugin.js
@@ -17,8 +17,8 @@ const JEST_TEST_FUNCTIONS = ["test", "it", "fit", "xit"];
  */
 function setupJestNockBack() {
   for (const jestTestFunction of JEST_TEST_FUNCTIONS) {
-    global[jestTestFunction].nock = function (testName, testFunction) {
-      global[jestTestFunction](testName, withNockBack(testFunction));
+    global[jestTestFunction].nock = function (testName, options, testFunction) {
+      global[jestTestFunction](testName, withNockBack(options, testFunction));
     };
   }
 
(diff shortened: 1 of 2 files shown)


loader/test/support/nock-back-plugin.js
```diff
@@ -59,21 +59,49 @@ function setupNockBack(mode, fixturePath) {
...
-function withNockBack(testFunction) {
+function withNockBack(options, testFunction) {
+  if (typeof options === "function") {
+    testFunction = options;
+    options = {};
+  }
+
+  const originalMode = nock.back.currentMode;
   module.exports.setupNockBack();
+  nock.back.setMode("record");
```
Mr0grog, line 80: This feels a little duplicative, so maybe the `setMode` part should come out of `setupNockBack`, or some other solution is needed here.

The issue is that, it turns out being in "record" mode can cause problems for other tests that use normal Nock instead of Nock.Back. My solution here was to turn "record" mode on and off at the start and end of each nock-back test. Not the prettiest but it works.

loader/test/support/nock-back-plugin.js
```diff
@@ -59,21 +59,49 @@ function setupNockBack(mode, fixturePath) {
...
  * responses if the upstream server has changed.
  *
- * @param {string} name The test name
+ * @param {any} [options] Options for handling recorded data.
+ * @param {boolean} [options.ignoreQuery] Ignore querystrings in requests.
+ * @param {(any) => void} [options.prepareScope] Function that will be called
+ *        with each recorded request definition before installing them to be
+ *        replayed. Use this, for example, to make recordings match when parts
+ *        of a request (e.g. the path) are random or change from test run to
+ *        test run.
```
Mr0grog, line 68: The Albertsons source (and, IIRC, some others) set random querystring variables to bust caches. This, of course, makes replaying recordings hard! `ignoreQuery` is entirely my own invention here, and causes querystrings in the requests that the test makes to be ignored entirely. It’s just a special case of `prepareScope`, which itself is a *very underdocumented* Nock-back feature. It’s a callback that lets you mess with the recorded data before setting it up for replay.

astonm: Makes sense to add this functionality, and I'm surprised it isn't already in Nock-back.

One slightly generalization that you might want to consider is allowing `ignoreQuery` to take a list of specific query parameters to remove, in case any of the others are meaningful.

astonm (approved):

PR #443, opened 2021-11-04
Expand store support in CDC API source

### LINKED ISSUE #393: Expand CDC loader to more store types and to non-stores
In #312, we implemented the CDC data loader to only load data from a subset of stores we knew were well supported and that the CDC provides “store numbers” for. For example, we filter locations for store name patterns: https://github.com/usdigitalresponse/univaf/blob/537ec0f8c249ee2f835d1648a6921ed4d8314d4d/loader/src/sources/cdc/api.js#L142-L152

We should work to expand our support for the CDC’s dataset to cover:
- More store chains. For example, we don’t include Shoprite, although they are in CDC’s data.
- Non-chain stores/pharmacies/clinics/hospitals. We don’t support these at all right now. In some cases, we can match up using our final snapshot data from Vaccinate The States (see #189 and #304).

Mr0grog:
This fixes Walmart/Sam's Club support in the CDC source, and also adds support for:
- Shoprite
- Stop & Shop
- Sav-On
- Genoa Healthcare
- Bartell Drug
- Meijer
- Winn Dixie
- Hannaford

I focused on chains that are common in our partners' states (AK, NJ, WA) and that had lots of entries in the CDC data. I also did my best to make sure these external IDs seem sensible and match up to what we’ve already got (when we already have something for these stores). One exception here is Wegmans, which I couldn’t find a reasonable way to match up and wrote a big comment about inline.

I also tried to do this with minimal impact to the architecture of this source, but having this confined to the scope of a single external ID poses some problems, like store types that should really set more than one, or that ought to change the location's name instead of just an external ID, or that reformat the store number/ID entirely but can't touch other fields in the output that use it (Walmart, with its `10-` prefix on the store number, is a good example). It might be worth refactoring this more in [the style of how VaccineSpotter was done](https://github.com/usdigitalresponse/univaf/blob/ca56bccbd2a7f66a6abe68595c99c442fc9f3440/loader/src/sources/vaccinespotter/index.js#L242-L404). But not sure it’s worth holding up this functionality for. 🤷 

Fixes #393.

### DIFF
--- a/loader/src/sources/cdc/api.js
+++ b/loader/src/sources/cdc/api.js
@@ -211,6 +211,28 @@ const locationSystems = [
   { system: "walgreens", pattern: /^Walgreens/i },
   { system: "sams_club", pattern: /^Sams Club/i, getId: getWalmartId },
   { system: "walmart", pattern: /^Walmart/i, getId: getWalmartId },
+  {
+    system: "shoprite",
+    // "Klein's Shoprite" is the same Shoprite as Shoprite, but just encompasses
+    // a subset of stores in Maryland.
+    pattern: /^(klein )?shoprite/i,
+    getId(location) {
+      let id;
+      // This is the only one not formatted with a # sign.
+      // TODO: this appears to be the same as store 322046, and I can't figure
+      // out which is "correct". Ideally we'd just put in both, but the current
+      // framework here can't handle that.
+      if (location.loc_name.toLowerCase().trim() === "Shoprite Pharmacy 801") {
+        id = "801";
+      }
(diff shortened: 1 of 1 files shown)


Mr0grog:
Thought for the future on Wegmans: Wegmans doesn't have multiple locations in the same physical place, so we could do geohashes, probably with a precision of 7 (roughly 150x150 meter box). We'd have to scope it to Wegmans (maybe name the system `wegmans_geohash`).

Mr0grog:
And name the system CDC has `wegmans_rx`.

Mr0grog:
Merging this for now; will address Wegmans and general refactoring stuff separately.

PR #446, opened 2021-11-05
Make CI workflow work from forks

Mr0grog:
This removes pointless secrets (the test database is a test database and lives only for the lifetime of the workflow; there is no reason for it to have a secret password) and disable ECR login & docker push commands from even trying to run anywhere except `main` (which implicitly excludes PRs).


Mr0grog:
# 🎉

Issue #459, opened 2021-11-15
Handle duplicate locations in PrepMod


Mr0grog:
Despite PrepMod's attempt to collapse multiple clinics at the same physical location into one entry, we’re seeing a lot of duplicate locations in practice. For instance, at the time of this writing, checking https://myhealth.alaska.gov/api/smart-scheduling-links/locations.ndjson gets a list of locations where the first two show up twice (with slightly differently worded addresses, which suspect is the problem). I didn’t check past the first two, but that’s not a great sign.

Given that we don’t expect geographically overlapping locations from PrepMod, using a geohash with precision 7 or 8 would probably be a good solution here.

We probably need to solve this in three ways:
1. Group found locations in the API response by geohash and format those *groups* into location entries (kind of the CDC API groups rows to be formatted together).
2. Add the geohash as an external ID (probably the system `prepmod-<host>-geo`) so we can match updates based on it in the server.
3. Do a one-off script or migration to add the above external IDs to existing prepmod locations in the DB.

Mr0grog:
Worthwhile geohash packages:
- `ngeohash` https://www.npmjs.com/package/ngeohash (Most broadly used, also has types through typeshed.)
- `@geonet/geohash` https://www.npmjs.com/package/@geonet/geohash (Nicer docs, with a precision reference! No types, though, and less used.)

Mr0grog:
Dug into this some more, and it does the wrong thing too often in our production data. Sometimes it appears to be bad geocoding in the underlying data, and sometimes there really are different places extremely close to (or even on top of!) each other, which we *thought* would be less true if we restricted things to PrepMod only. We were wrong. (The most common situation is different mobile clinics setting up at the same place on different days, which is definitely a messy situation as far as our data model goes. In any case, it’s certainly not obvious that we should merge them.)

I don’t think there’s really a technical solution here, although there is probably some manual one-off deduping we can do.

Mr0grog:
Did some manual cleaning up and looking around. Filed #842 to fix a minor merge issue. There is definitely just a lot of messy data with Washington’s PrepMod instance (less so with Alaska 🤷).

[...]

Issue #458, opened 2021-11-15
Some PrepMod locations have full addresses in `address.line`


Mr0grog:
Some PrepMod locations have full addresses in the `address.line` field. For example:

```js
{
  "id": "52be689dfd69754fdaae719fb8805354",
  "resourceType": "Location",
  "address": {
    "line": [
      "Loussac Library 3600 Denali St, Anchorage, AK 99503"
    ],
    "city": "Anchorage",
    "state": "Alaska",
    "postalCode": "99503",
    "district": "Anchorage"
  },
  // Other fields...
}
```

Where possible we should remove the `, Anchorage, AK 99503` on the end. One way might be to do something like the following to the last entry in the `address.line` array:

```js
line.replace(/, [A-Za-z\s]+, [A-Z]{2}, \d{5}(-\d{4})?$/, "")
```

This might also be something worth applying to other loaders, although it seems to be a particularly prevalent issue in PrepMod.

Issue #449, opened 2021-11-06
Mark locations as private if they disappear from source data


Mr0grog:
We sometimes have locations that were once available for vaccination, but no longer are. It’s hard to know when a place disappears in any automatic way, but if we know a source has complete coverage (e.g. CVS’s SMART SL API should reliably list every CVS location that is doing vaccinations), we could determine this by comparing what the source sends with locations we already have in the database.

For example, the CVS SMART source might:
1. Request `https://getmyvax.org/api/edge/locations?provider=cvs` and save that list in memory.
2. Do its work as normal, but for each location it finds in the source data, remove it from the list created in step 1.
3. Output an update for every location still remaining in the list from step 1, but with the `is_public` property set to `false`.

Right now, this is most needed for Prepmod — in Alaska, there have been several locations accidentally published publicly and then later made private, but we still show it because we picked it up when it was briefly public. However, it’s still relevant to any other source that is canonical in some way (e.g. Walgreens/CVS/Kroger SMART, Albertsons).

- [x] PrepMod #460
- [ ] Walgreens
- [ ] CVS
- [ ] Rite Aid
- [ ] Kroger
- [ ] Albertsons
- [ ] HyVee

brendanbabb:
There was a clinic that was public for our mass shelter that was supposed to be private, named Visit Health Mobile Clinics (Address: Sullivan Arena 1600 Gambell St.) We contacted the Anchorage health dept. to change the PrepMod clinic from a public to private clinic. We still see the clinic in the univaf API but not in the PrepMod API. We have seen a similar bug twice before.

We just had the clinic changes from Public to Private around 12:30pm, 11/10/2021

The clinic is for Sullivan Arena and here is the PrepMod link: https://myhealth.alaska.gov/appointment/en/client/registration?clinic_id=5068

univaf_url <- "https://getmyvax.org/api/edge/locations.ndjson?source=univaf-prepmod&state=AK" (Showing up in this URL)
univaf_url_2 <- "https://getmyvax.org/api/edge/locations.ndjson?source=univaf-prepmod&state=Alaska"

The clinic is now private but is showing up in the Univaf API under id: 76b6d85e-dc2b-413e-9b9f-3ea4f02f2072

It is not showing up in the PrepMod API: locations_url <-  "https://myhealth.alaska.gov/api/smart-scheduling-links/locations.ndjson"

Please let us know if you need additional info.

Mr0grog:
[...]

PR #460, opened 2021-11-15
Hide locations that disappear from the PrepMod API

Mr0grog:
We’ve had a lot of instances where a location in Alaska managed by PrepMod may initially be public, but is later marked private. This solves the problem by comparing the list of locations we already know about with the list of locations found in the PrepMod API, and marking any previously known locations as private if they didn’t show up in the PrepMod API.

This addresses #449 for PrepMod. It still needs to be propagated to other sources where it applies (CVS, Walgreens, Rite Aid, etc.).

A few other fixes came along for the ride here:
- `ApiClient.getLocations()` wasn’t respecting pagination! I think we used it in pretty constrained scenarios before (NJVSS) and never expected too many results. Honestly the situation is similar for PrepMod, but this felt worth fixing.
- Use the same `warn(error, context)` pattern for PrepMod that we use in many of the other sources. This still needs to be more streamlined across all sources, but just adding it here improved things a bit.
- Stop warning about Flu vaccine info in PrepMod; we should be expecting it, even if we aren’t recording it.

I also discovered a couple other PrepMod issues that I didn’t fix here because they felt too big, but want to return to immediately after this:
- #458
- #459

### DIFF
--- a/loader/src/api-client.js
+++ b/loader/src/api-client.js
@@ -22,30 +22,39 @@ class ApiClient {
     this.key = key;
   }
 
-  async getLocations(query) {
-    let options = {
-      url: `${this.url}/api/edge/locations`,
-      searchParams: query,
-      headers: { "x-api-key": this.key },
+  async _request(options) {
+    return httpClient({
+      ...options,
+      url: new URL(options.url, this.url).href,
+      headers: { "x-api-key": this.key, ...options.headers },
       responseType: "json",
-    };
+    });
+  }
+
+  async _requestAllPages(requestOptions) {
+    // This gets updated on each response with the options for the next page.
+    let nextOptions = requestOptions;
 
     const results = [];
-    while (options) {
-      const { body } = await httpClient(options);
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
+      const { body } = await this._request(nextOptions);
       results.push(...body.data);
 
       if (body.links?.next) {
-        options = {
-          ...options,
-          url: new URL(body.links.next, this.url).href,
-          searchParams: undefined,
-        };
+        nextOptions = { url: body.links.next };
       } else {
-        options = null;
+        break;
       }
     }
     return results;
+  }
+
+  async getLocations(query) {
+    return this._requestAllPages({
+      url: "/api/edge/locations",
+      searchParams: query,
+    });
   }
 
   async sendUpdate(data, options) {
@@ -53,12 +62,11 @@ class ApiClient {
       throw new TypeError("`options` must be an object");
     }
 
-    const response = await httpClient.post({
-      url: `${this.url}/api/edge/update`,
+    const response = await this._request({
+      method: "POST",
+      url: "/api/edge/update",
       searchParams: options,
-      headers: { "x-api-key": this.key },
       json: data,
-      responseType: "json",
       throwHttpErrors: false,
     });
 
--- a/loader/src/api-client.js
+++ b/loader/src/api-client.js
@@ -23,13 +23,29 @@ class ApiClient {
   }
 
   async getLocations(query) {
-    const { body } = await httpClient({
-      url: `${this.url}/locations`,
+    let options = {
+      url: `${this.url}/api/edge/locations`,
       searchParams: query,
       headers: { "x-api-key": this.key },
       responseType: "json",
-    });
-    return body;
+    };
+
+    const results = [];
+    while (options) {
+      const { body } = await httpClient(options);
+      results.push(...body.data);
+
+      if (body.links?.next) {
+        options = {
+          ...options,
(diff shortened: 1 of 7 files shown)


loader/src/api-client.js
```diff
@@ -23,13 +23,29 @@ class ApiClient {
...
+      url: `${this.url}/api/edge/locations`,
       searchParams: query,
       headers: { "x-api-key": this.key },
       responseType: "json",
-    });
-    return body;
+    };
+
+    const results = [];
+    while (options) {
```
astonm, line 34: Any reason not to make this `while (true) {`...

loader/src/api-client.js
```diff
@@ -23,13 +23,29 @@ class ApiClient {
...
+      results.push(...body.data);
+
+      if (body.links?.next) {
+        options = {
+          ...options,
+          url: new URL(body.links.next, this.url).href,
+          searchParams: undefined,
+        };
+      } else {
+        options = null;
```
astonm, line 45: ...and then `break` here instead? Code as written works, but it took me a little to understand the overloading of `options` as both the configuration for the next request and also a flag for continuing the loop.

If you do prefer it this way, maybe naming it `nextOptions` would aid in signaling it's being prepared for use elsewhere and that once `null` it means there's nothing to do next.

loader/src/sources/prepmod/index.js
```diff
@@ -41,6 +52,27 @@ async function getDataForHost(host) {
...
 
+async function getKnownLocations(state) {
+  const client = ApiClient.fromEnv();
+  const locations = await client.getLocations({
+    state,
+    provider: "prepmod",
+  });
+
+  // Create a lookup object indexed by external ID.
+  const result = new Map();
```
astonm, line 63: Is the reason to use a `Map` here rather than a plain old object just because you wanted `.values()` below? Seems like convention elsewhere in the codebase is to call `Object.values(...)`

loader/test/prepmod.test.js
```diff
@@ -1,15 +1,28 @@
...
 const { EXTENSIONS } = require("../src/smart-scheduling-links");
 
 describe("PrepMod API", () => {
   jest.setTimeout(60000);
 
+  const _apiUrl = config.apiUrl;
+  const _apiKey = config.apiKey;
   afterEach(() => {
+    config.apiUrl = _apiUrl;
+    config.apiKey = _apiKey;
```
astonm, line 25: Don't love this hack. Do we have to mutate these globals?

Mr0grog: We do, since `checkAvailability` will read the config when it sets up an `ApiClient` instance. I might also be able to use [`jest.mock(moduleName, newModuleDefinition)`](https://jestjs.io/docs/jest-object#jestmockmodulename-factory-options), but IIRC it has significant caveats when mocking properties and not functions. I remember having some issues with it around simple stuff like this before.

loader/src/sources/prepmod/index.js
Mr0grog: No special reason, but I guess I tend towards `Map` when it isn’t especially more or less useful than a plain object, since offers the same API when you need its more special features as when your indexes are strings. I guess it’s fair to say we’ve used plain objects in a lot of other places I normally wouldn’t, so this feels out of place as a result. Can change.

loader/src/api-client.js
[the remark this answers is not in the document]
Mr0grog: Nah, not a big reason. This is one of those ”mentors early on pounded into me to avoid `while(true)` whenever possible so it’s just what I tend to do” things. (Unsurprisingly, during a time when I wrote more C code. 😜)

Will change.

loader/test/prepmod.test.js
Mr0grog: If you don’t think, say, anything but `cli.js` should ever use the `config.js` module, that’s not crazy, but it does mean we have to refactor everything, which I think is a little out of scope here. 😉

The other option here be just always setting these values to something arbitrary like `"http://univaf.test"` and `"test-api-key"` for all tests, but I was trying to keep it narrow here. We also still need to do this kind of thing in general for other tests that actually test how behavior differs with different configuration, e.g. https://github.com/usdigitalresponse/univaf/blob/fddfe0a71dd187be8bba7e273c5cd59f26f4e769/loader/test/cvs.api.test.js#L22-L39 (although the `config` module never got backported to that source [inherited from New Jersey], it's the same principle).

astonm: I don't mind at all having one big override for all of the tests, if that would clean up this PR. But I don't think it's too big of a deal either way. Just struck me as odd.

astonm (approved):
Changes look good to me. 🚢

PR #467, opened 2021-11-18
Provide a best-effort booking URL for PrepMod locations based on clinic search criteria

Mr0grog:
PrepMod provides an info URL that simply links to the front page of the PrepMod server, which includes *many* more locations than the one in question, or provides ultra-granular links to individual booking slots, which can be a bit hard for consumers to manage. In addition, it turns out some clinics have complicated vaccination patterns, like one in Alaska where Moderna is offered on certain days and J&J on others, but where the PrepMod API does not provide separate schedules for the two vaccines, so the product information on the slots is inaccurate and unhelpful.

This attempts to provide a happy medium by making the `booking_link` property on a location (the slot one is left as-is) that links to search results for the clinic. The caveat here is that searches are geographically based on the centroid of a zip code, and so may not include all of a zip code or may include other zip codes. A big radius is not a problem when a location's name is unique (e.g. "Bernie's Pharmacy"), but is tough to deal with when the name is generic or refers to a mobile clinic (e.g. "Anchorange Public Health Mobile Clinic"). We use a 10 mile radius as a happy medium. We might need to adjust in the future.

### DIFF
--- a/loader/src/sources/prepmod/index.js
+++ b/loader/src/sources/prepmod/index.js
@@ -116,6 +116,7 @@ function formatLocation(host, validTime, locationInfo) {
     county: smartLocation.address.district || undefined,
     position,
 
+    booking_url: formatLocationBookingUrl(host, smartLocation),
     info_url,
     info_phone,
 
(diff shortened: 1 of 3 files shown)


Mr0grog:
Going ahead and merging this so it’s available for Alaska.

PR #473, opened 2021-11-23
Support new pediatric data in Albertsons

Mr0grog:
Albertsons started showing pediatric data (yay!) but they did it by having separate records for adult vs. pediatric vaccines (messy for us). See the added test in this PR for a good example.

This attempts to handle that by grouping and merging locations with the same store number (or name, if lacking a store number) after formatting. (We could also group before formatting; it felt kind of like a tossup on complexity to me.)


loader/test/albertsons.test.js
```diff
@@ -206,6 +212,83 @@ describe("Albertsons", () => {
...
+    const result = await checkAvailability(() => {}, { states: "VA" });
+    expect(result).toContainItemsMatchingSchema(locationSchema);
+    expect(result).toEqual([
+      {
+        name: "Safeway 0005",
+        external_ids: [
+          ["albertsons", "1635993536219"],
+          ["albertsons_safeway", "1635993536219"],
+          ["safeway", "0005"],
+          ["albertsons_store_number", "safeway:0005"],
```
Mr0grog, line 257: It *seemed* useful to surface this `albertsons_store_number` field as an external ID, but it’s actually only used for identifying two locations that need merging. This could also be in `meta` (which would simplify some code), or hidden from the output entirely.

loader/test/albertsons.test.js
```diff
@@ -206,6 +212,83 @@ describe("Albertsons", () => {
...
+          checked_at: expectDatetimeString(),
+          valid_at: "2021-10-28T07:06:13.000Z",
+        },
+        meta: {
+          albertsons_region: "Virginia",
+          booking_url_adult:
+            "https://kordinator.mhealthcoach.net/vcl/1600100807144",
+          booking_url_pediatric:
+            "https://kordinator.mhealthcoach.net/vcl/1635993536219",
+        },
```
Mr0grog, line 287: I couldn’t think of a better approach to preserving this info than putting it in `meta`. Will have to notify Anchorage and NJ folks about it.

I wanted to keep the top-level `booking_url` field as a nice, deep link, but it now has the downside of only working for adult vaccines. Another option is to put a more general link like https://www.mhealthappointments.com/covidappt in the top level, so it works for all vaccine types.

loader/src/sources/albertsons/index.js
```diff
@@ -11,8 +11,13 @@
  */
 
 const Sentry = require("@sentry/node");
+const _ = require("lodash");
```
astonm, line 14: I think the bundle size may be smaller if you just `const groupBy = require("lodash/groupBy")`.

loader/src/sources/albertsons/index.js
```diff
@@ -206,18 +213,69 @@ async function getData(states) {
...
     })
     .filter((location) => states.includes(location.state));
+
+  // Adult & Pediatric vaccines at the same locations get different listings in
+  // the Albertsons API, so we need to group and merge them.
+  const groups = _.groupBy(formatted, (location) => {
+    return (
+      location.external_ids.find(
+        (id) => id[0] === "albertsons_store_number"
+      )?.[1] ?? location.name
```
astonm, line 223: `??`, razzle dazzle!

I will admit, it took me many re-reads to grok all this is doing. Since it's already multiline, maybe worthwhile being less terse?

```
const storeExtId = location.external_ids.find((id) => id[0] === "albertsons_store_number");
return storeExtId ? storeExtId[1] : location.name;
```

or, since you have lodash now

```
const storeExtId = location.external_ids.find(_.matches({0: "albertsons_store_number"}));
return storeExtId ? storeExtId[1] : location.name;
```

loader/src/sources/albertsons/index.js
```diff
@@ -206,18 +213,69 @@ async function getData(states) {
...
+        "Trying to merge locations other than an adult and pediatric!",
+        group
+      );
+    }
+
+    const result = Object.assign({}, ...group);
+    result.meta = {
+      ...adult.meta,
+      booking_url_adult: adult.booking_url,
+      booking_url_pediatric: pediatric.booking_url,
```
astonm, line 244: In the weird case where `!adult || !pediatric` one of these two lines will blow up. I think maybe you should just exit the function rather than `warn`ing above.

loader/src/sources/albertsons/index.js
Mr0grog: Good point. That said, I think this mainly looks confusing because of Prettier trying to shorten the lines, which still happens even if we separate them.  :\

```js
const storeId = location.external_ids.find(
      (id) => id[0] === "albertsons_store_number"
    );
    return storeId ? storeId[1] : location.name;
```

But it’s definitely clearer than what I’ve got now.

loader/src/sources/albertsons/index.js
Mr0grog: Oh, geez, I know I had a `return null` there at one point. Will fix!

loader/src/sources/albertsons/index.js
Mr0grog: I already started using `mapKeys` in #474, so I think it’s probably easier to pull in all of lodash as a dependency so expanding usage isn’t a pain (and we aren’t worried about running in a browser here). We’ll see if I regret that later, though.

astonm (approved):
👍🏽

Issue #474, opened 2021-11-23
Write scraper for Rite Aid


Mr0grog:
All of Rite Aid's APIs (the one we currently use, the one VaccineSpotter used to use, and the newer, public SMART SL API in #159) have been down for almost a week. We’ve spent much of that time trying to get in touch with them on various channels, along with our contacts at the State of NJ and USDS, but they’ve been completely unresponsive.

So! It’s time to start scraping, unfortunately.

It looks like their scheduling web app has its own API for use by XHR (yay!) and that we can hit at URLs like: `https://www.riteaid.com/content/riteaid-web/en.racloudgetavailablestores.json?address=08103&radius=25&vaccineIds=9-PREF-113,11-PREF-114,13-PREF-115&showAllStoresWithAvailibility=false&fetchMechanismVersion=2&count=100`

I can’t figure out how to page through results, but you can set `count` and `radius` pretty high, so that may eliminate any need for it.

`vaccineIds` appear to be:
- `9-PREF-113`: Moderna Adult Dose 1, or Pfizer Adult Dose 2 (but in this case add `&dose1appointmentDate=2021-10-22T00:00:00`)
- `42-PREF-113`: Moderna Adult Booster
- `11-PREF-114`: Pfizer Adult Dose 1, or Pfizer Adult Dose 2 (but in this case add `&dose1appointmentDate=2021-10-22T00:00:00`)
- `43-PREF-114`: Pfizer Adult Booster
- `13-PREF-115`: J&J Adult Dose 1
- `46-PREF-115`: J&J Adult Booster 1
- `47-PREF-122`: Pfizer Pediatric Dose 1

This is high priority, since there is actually a real crunch right now (booster and pediatrics), and locations often don’t have appointments available. It’s super important for NJ.

Mr0grog:
It looks like Washington’s API is not showing Rite Aid (e.g. https://vaccinelocator.doh.wa.gov/locations/98115), likely because of this problem, so once we’ve got things solved here, we should let them know they can use us if helpful.

Mr0grog:
This was solved in #488.

PR #475, opened 2021-11-23
Add scraper for Rite Aid

Mr0grog:
Why are we scraping Rite Aid? See #474.

This is based on zip code queries for 100 mile radiuses, and keeps a list of the zips that will fairly efficiently cover any state in `sources/riteaid/zip-codes.js`. I calculated these by hand for small states where one or two zips covers them, plus Washington, which is a priority for us. I got some pointers from a friend on calculating an efficient set of zips for larger states, which I'll try and incorporate on a second pass.


Mr0grog:
OK, I think this is basically good for a deployable first pass.

Mr0grog:
Err, *now* it is.

Mr0grog:
(Rebased after merging #473.)

astonm (approved):
Tight timeline, I know. Haven't had a chance to review in depth, but from a superficial pass definitely looks shippable.

Mr0grog:
Thanks! 🙏

PR #485, opened 2021-12-01
[364] Added ability to set image tag for loaders

jaronoff97:
This PR changes the loader TF module to allow us to set the version for the image. By doing this, we will be able to make the image repositories immutable to prevent any image tags from getting overwritten. That work will be covered in a follow up PR so as to isolate the changes being made.


jaronoff97:
The TF diff has to change the task definitions for all of the loaders, so we see thirteen changes in all.

Mr0grog:
Rebased so this merges cleanly.

terraform/modules/loader/main.tf
```diff
@@ -2,7 +2,7 @@ module "loader_task" {
   source = "../task"
 
   name    = var.name
-  image   = "681497372638.dkr.ecr.us-west-2.amazonaws.com/univaf-loader"
+  image   = "${var.loader_image}"
```
Mr0grog, line 5: Let’s fix the deprecation warning this triggers:

```
Warning: Interpolation-only expressions are deprecated

  on modules/loader/main.tf line 5, in module "loader_task":
   5:   image   = "${var.loader_image}"
```

Mr0grog (changes requested):
Sorry I missed this was waiting on me! Looks good, except for the deprecated syntax. Fix that and let’s get this in!

Mr0grog:
Oh, do we also need to stop pushing a `latest` tag here? https://github.com/usdigitalresponse/univaf/blob/71bf9a488a00d5f80e4b2737d61a2901bd9ebd4e/.github/workflows/ci.yml#L194-L202

Mr0grog (approved):
👍 👍 👍 Merge at will :shipit: :shipit: :shipit:

PR #487, opened 2021-12-03
Alaska has shingles (vaccines)

Mr0grog:
This got a lot more complex than expected. Alaska is now including a shingles (zoster) vaccine in PrepMod, so we need to treat it like flu vaccines and ignore it.

While writing a test for that, I realized we weren’t correctly handling locations that only had non-COVID vaccines — we’d still include them, and set `availability` based on the non-COVID slots rather than COVID slots! That’s obviously pretty wrong. So this does a little bit of refactoring to determine whether a schedule is not actually a COVID schedule and skip over it (and its slots) if not, and drop the location if all its schedules are non-COVID ones.

### DIFF
--- a/loader/src/sources/prepmod/index.js
+++ b/loader/src/sources/prepmod/index.js
@@ -47,9 +47,9 @@ async function getDataForHost(host) {
   const api = getApiForHost(host);
   const manifest = await api.getManifest();
   const smartLocations = await getLocations(api);
-  return Object.values(smartLocations).map((entry) =>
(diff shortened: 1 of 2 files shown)


astonm (approved):
Complicated, but looks fine to me.

👍🏽

Mr0grog:
FWIW, I think the indentation change makes it look bigger than it really is, but yeah. :\

If I ever have time to get around to it, I think I should probably make classes for the SMART SL objects (e.g. schedules, slots) that incorporate some of this logic (e.g. with an `isCovid` property/method).

PR #488, opened 2021-12-04
Add remaining zip codes for Rite Aid Scraping

Mr0grog:
Thanks to @tmcw who helped with some code to narrow down a complete list of zip codes into a smaller list of fitting candidates.

I also discovered that Rite Aid has closed a few stores over the last couple months, including *every store in Colorado:* https://www.riteaid.com/locations/co/grand-junction.html (Also I did a lot more work than necessary here, assuming there were plenty more stores I’d need to account for before discovering there were only 3. 🤦)

### DIFF
--- a/loader/src/sources/riteaid/scraper.js
+++ b/loader/src/sources/riteaid/scraper.js
@@ -92,8 +92,11 @@ async function* queryState(state, rateLimit = null) {
     if (rateLimit) await rateLimit.ready();
 
     const body = await queryZipCode(zipCode);
-    for (const item of body.data.stores) {
-      if (!seenStores.has(item.storeNumber)) {
+
+    // If there are no results, `body.data.stores` is `null`.
+    const stores = body.data.stores || [];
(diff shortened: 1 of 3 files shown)


loader/src/sources/riteaid/scraper.js
```diff
@@ -92,8 +92,11 @@ async function* queryState(state, rateLimit = null) {
     if (rateLimit) await rateLimit.ready();
 
     const body = await queryZipCode(zipCode);
-    for (const item of body.data.stores) {
-      if (!seenStores.has(item.storeNumber)) {
+
+    // If there are no results, `body.data.stores` is `null`.
+    const stores = body.data.stores || [];
+    for (const item of stores) {
+      if (!seenStores.has(item.storeNumber) && item.state === state) {
```
Mr0grog, line 99: We’ve been seeing the occasional error like this, and I finally discovered why! I don’t understand why it’s intermittent, though.

PR #495, opened 2021-12-13
Remove bad PrepMod location data from WA DoH

Mr0grog:
It looks like WA DoH has some bad PrepMod data in their DB. This fixes the issue by:
- Adding the location as to our list of bad data exceptions.
- Skipping over PrepMod-sourced data entirely when importing from WA DoH (since we already get it direct from PrepMod).
- Using a migration to hide the bad data in our DB.

### DIFF
--- a/loader/src/sources/wa-doh.js
+++ b/loader/src/sources/wa-doh.js
@@ -66,6 +66,14 @@ const LOCATIONS_QUERY = `
   }
 `;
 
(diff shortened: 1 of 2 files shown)


Mr0grog:
Once this is merged, I’ll also try to get in touch with WA DoH folks and let them know about the bad location record. (It looks like we never imported it from PrepMod ourselves, so maybe this is an error in their code. 🤷)

astonm (approved):
👍🏽

PR #516, opened 2022-01-06
Albertsons data just gets funkier and funkier

Mr0grog:
This makes two changes to resolve issues in Albertsons data right now:

1. Skip over locations with obviously invalid addresses (e.g. `., ., TX 75244`). There are two of these right now, and they currently parse “successfully” because they have an address-like structure. I’m betting they don’t refer to an actual location (the name for both of them is “Dallas ISD” which is probably the Dallas Independent School District), and if someone tried to book them, it’s not really clear where they would actually go (clicking through to the booking screen doesn’t provide any more useful information to clarify what or where this location is supposed to be). This will still emit warnings for bad addresses, but at least we won’t surface them in final data, and the issue is clearer.

2. Add manual corrections for a Jewel-Osco location (and a framework for future corrections). This location doesn’t have correct-enough data to “fix” it algorithmically like we do in some other locations (e.g. with repeated store names). Supporting manual corrections was the best I could think of here.

Fixes these two issues in Sentry right now:
- https://sentry.io/organizations/usdr/issues/2829716008 (1 above — different locations getting conflated because of bad addresses)
- https://sentry.io/organizations/usdr/issues/2809052505 (2 above — manual corrections)

### DIFF
--- a/loader/src/sources/albertsons/corrections.js
+++ b/loader/src/sources/albertsons/corrections.js
@@ -1,0 +1,10 @@
+/**
+ * Manual corrections for misformatted/bad source data. The keys are the `id`
+ * property for the location, and the value is any properties that should be
+ * overridden.
+ */
+module.exports.corrections = {
+  1641420736546: {
+    address: "Jewel-Osco 3441 - 2940 N Ashland Ave, Chicago, IL, 60657",
+  },
+};
--- a/loader/src/sources/albertsons/index.js
+++ b/loader/src/sources/albertsons/index.js
@@ -33,6 +33,7 @@ const {
   PediatricVaccineProducts,
 } = require("../../model");
 const { ParseError } = require("../../exceptions");
+const { corrections } = require("./corrections");
 
 const API_URL =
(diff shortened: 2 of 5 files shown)


Mr0grog:
Also open to suggestions as to whether there is a good or safe-seeming way to totally silence the Dallas ISD locations, since this fix will still capture warnings in Sentry (they’ll just be better and clearer, and we’ll know to ignore them in this particular case).

astonm (approved):
Overall looks good to me. One tweak that I probably would recommend (although, run it against the dataset to see if it's sane) is to require that the address starts with a number.

Mr0grog:
Turns out we have 404 addresses that violate that constraint, so I’m going to skip that one. (Some probably would have benefitted from it, e.g. “Vons - 4627 Carmel Mountain Road”, but others are definitely not compatible in actuality, e.g. “PHKC-CE 4520 Martin Luther King Jr Way S” or “Walla Walla County Fairgrounds” or, [wow] “Inside Olympia Capital Mall, next to Buckle, near Shoe Dept. Encore - 625 Black Lake Boulevard Southwest.”)

PR #521, opened 2022-01-11
Send/store unpadded store numbers for Albertsons

Mr0grog:
While cleaning up duplicate Walmarts in #519, I discovered some duplicate zero-padded/unpadded Albertsons that appear to have been caused by us not unpadding external IDs we send from the Albertsons loader (it turns out some entries are zero-padded and some not, and sometimes the pediatric and non-pediatric entries for the same location have different padding).

### DIFF
--- a/loader/src/sources/albertsons/index.js
+++ b/loader/src/sources/albertsons/index.js
@@ -254,7 +254,7 @@ async function getData(states) {
     const storeId = location.external_ids.find(
       (id) => id[0] === "albertsons_store_number"
     );
-    return storeId ? unpadNumber(storeId[1]) : location.name;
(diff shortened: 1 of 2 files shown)


loader/src/sources/albertsons/index.js
```diff
@@ -254,7 +254,7 @@ async function getData(states) {
     const storeId = location.external_ids.find(
       (id) => id[0] === "albertsons_store_number"
     );
-    return storeId ? unpadNumber(storeId[1]) : location.name;
+    return storeId?.[1] ?? location.name;
```
Mr0grog, line 257: `unpad()` is no longer needed here since we’ve now guaranteed that the data is already unpadded when parsing.

PR #531, opened 2022-01-14
Keep trying additional states if Rite Aid API fails on one

Mr0grog:
We have an issue where the Rite Aid API loader will often have an error on a request for a state’s data and stop, instead of continuing on to try the rest of the states. This handles that, and sends better error info to Sentry to boot.

It also adds rate limiting support.


Mr0grog:
It looks like the underlying problem here is that our API token has been disabled for Colorado. These are still good changes to make, though.

PR #532, opened 2022-01-14
Load slot-level detail in second pass for Rite Aid scraper

### LINKED ISSUE #528: Rite Aid Scraper is currently showing all locations as unavailable
It looks like the Rite Aid data format changed so `availableSlots` is always an empty array. Need to figure out what different API calls we need to make to get actual data.

Mr0grog:
The Rite Aid web UI now makes *two* requests for each zip code — the first one returns a list of stores, in *mostly* the same format as before. The difference is that the array of slots in it is now empty and it has a `firstAvailableSlot` property. You then have to make a second call to the same endpoint, but with a list of store numbers in the querystring, to get slots filled in.

The first two changes make the code able to handle both the new summary/non-slot data and to optionally make extra requests for the slot-level data. (It gets the slot data by default, but it’s a function argument so we can easily turn it off if needed.)

The next change adds fairly strict validation of the returned data with AJV. My thinking here is to avoid a subtle change like the one that precipitated this — we’d get notifications that `firstAvailableSlot` was added, and would know to take a look at what’s going on. (That said, this one would still be subtle to understand the important of — `availableSlots` is still there, and still an array, it’s just empty. But I think this is the best we can do for proactive notification.)

Fixes #528.


Mr0grog:
This is a work-in-progress. I need to clean up the AJV validation code (the schema is good, but it needs a light-weight wrapper to make setup and throwing exceptions nicer).

Mr0grog:
OK, should be ready to go.

astonm (approved):
Wow, kind of a pain. Code looks good, though. Hope we use the `schema-validation.js` code lots going forward.

Mr0grog:
Yeah, in hindsight, we should definitely have been doing something like this more.

PR #568, opened 2022-02-16
Switch to use new aws_s3_bucket_acl config style

Mr0grog:
It looks like the default version of the AWS provider for Terraform was updated since our last deploy, and there were breaking changes (there may have deprecation warnings we missed before; I’m not really sure). The `acl` property was made read-only, so now we need to use a more verbose syntax to apply the same standard "read-only" ACL to our bucket. See also https://registry.terraform.io/providers/hashicorp%20%20/aws/latest/docs/resources/s3_bucket#using-acl-policy-grants

### DIFF
--- a/terraform/provider.tf
+++ b/terraform/provider.tf
@@ -8,6 +8,6 @@ terraform {
 }
 
 provider "aws" {
-  profile = "default"
+  profile = ""
   region  = var.aws_region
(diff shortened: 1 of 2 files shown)


Mr0grog:
This probably needs to be re-done as part of #571.

PR #570, opened 2022-02-16
Set up Terraform dependencies correctly

Mr0grog:
This is a follow-on to #569. It turns out we were using an old method of configuring Terraform dependencies and versions that is deprecated. This updates it and adds Dependabot for for Terraform and actions (but not for NPM, where we are still using Snyk for now).


Mr0grog:
This successfully clears the warnings on Terraform cloud!

PR #598, opened 2022-03-15
Ensure locations’ `state` is a two letter postal abbreviation

### LINKED ISSUE #444: Handle invalid state abbreviations from PrepMod
Some locations in the database have state names instead of postal abbreviations, e.g. "Alaska" instead of "AK" (it's possible there are other errors and misspellings, too). We need to do some validation and correction of these, otherwise it's tough for our state partners to query the API reliably.

I’m pretty sure these are all coming from PrepMod right now (they previously had a bug with this on their side that was supposedly fixed, so there might have been a regression).

We could fix it directly in the PrepMod loader or more generically in the server during data validation/correction, which might be better.

Mr0grog:
This adds some validation logic to the server that ensures we always store states in the form of a two-letter postal abbreviation (e.g. “AK” for Alaska). This is already what we mostly have, but there are a handful of locations that have come in from PrepMod with different names (e.g. “alaska” or “Alaska”; PrepMod appears to allow free-form data in the state field). The basic premise here is to accept any known identifier for the state (it‘s full name, a postal abbreviation, an ISO name, etc.) and replace that with the postal abbreviation, or return a 422 error if we don’t recognize the state that was passed in.

While I was at it, I went ahead and wrote a more generalized `validateLocationInput()` function that does a few other minor validations (it handles one we already had, and adds a few others for safety). It stops short of validating the complete schema of the location input object, but I’ll add that in if anyone thinks we should just go ahead and do it.

This also includes a migration to clean out the existing bad data.

Fixes #444.

### DIFF
--- a/server/src/validation.ts
+++ b/server/src/validation.ts
@@ -164,6 +164,61 @@ const CAPACITY_SCHEMA = {
   ],
 };
 
+const PROVIDER_LOCATION_SCHEMA = {
+  type: "object",
+  properties: {
+    // This should normally have `format: "uuid"`, but the server will accept
+    // old-style IDs (which were not UUIDs) and handle them specially.
+    id: { type: "string", nullable: true },
+    external_ids: {
+      type: "array",
+      nullable: true,
+      items: {
+        type: "array",
+        items: { type: "string", minLength: 1 },
+      },
+    },
+    provider: { type: "string" },
+    // We have an enum for location type, but because it's separate from the
+    // loader codebase, keep the validation loose. (Ideally we should have
+    // some real code sharing between the two.)
+    location_type: {
+      type: "string",
+      transform: ["toUpperCase"],
+      nullable: true,
+    },
+    name: { type: "string" },
+    address_lines: { type: "array", items: { type: "string" }, nullable: true },
+    city: { type: "string", nullable: true },
+    state: { type: "string" },
+    postal_code: { type: "string", nullable: true },
+    county: { type: "string", nullable: true },
+    position: {
+      type: "object",
+      nullable: true,
+      properties: {
+        longitude: { type: "number", format: "float" },
+        latitude: { type: "number", format: "float" },
+      },
+    },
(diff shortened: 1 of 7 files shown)


server/src/validation.ts
```diff
@@ -316,3 +318,65 @@ export function validateAvailabilityInput(data: any): AvailabilityInput {
...
+  if (requiredFields) {
+    const missing = ["name", "provider", "state"].filter(
+      (field) => !result[field]
+    );
+    if (missing.length) {
+      throw new ValueError(
+        `The location is missing values for: ${missing.join(", ")}`
+      );
+    }
+  }
```
Mr0grog, line 379: I got a bit lazy here, I guess. Not sure if I should go ahead and add a schema object for this and validate that. (Somewhat complex because we need two — one that requires some fields and one that doesn’t, since updates don’t have any real required fields.)

astonm: Sure, not clear you need to go through a lot of effort here to validate things further.

But if you did want to, just curious, any easy way to construct a derivative or child validator such that you could make the looser one and then overlay the required fields on top for the second?

astonm (approved):
👍🏽 

All looks good to me.

server/src/validation.ts
```diff
@@ -316,3 +318,65 @@ export function validateAvailabilityInput(data: any): AvailabilityInput {
...
+      "iso",
+      "usps",
+      "gpo",
+      "ap",
+    ];
+    for (const state of states) {
+      for (const key of references) {
+        const reference = state[key] as string;
+        if (reference) {
+          stateLookup.set(reference.toLowerCase(), state);
```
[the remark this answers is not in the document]
Mr0grog: Well I’m glad you suggested that. Certainly fixed some subtle errors, and was more complex than I thought!

server/src/validation.ts
Mr0grog: Went ahead and did this. Hopefully won’t cause any issues on real data! 🤞

PR #603, opened 2022-03-18
Upgrade Node.js to v16.14.2 (latest LTS version)

Mr0grog:
This upgrades us from v14 to v16 of Node.js, which adds some nice new JS/TS language features. I’ve also switched to `-slim` Docker images, but have not really tested out that they work yet. Need to do that before merging.


Mr0grog:
Verified `-slim` seems to work well, at least locally. (Wow, been a while since I've developed this with docker.)

PR #607, opened 2022-03-19
Repair not-quite-valid URLs from CDC data

Mr0grog:
In #598 we added more robust schema checking for location data, and it turns out we were importing lots of not-quite-valid URLs from the CDC's dataset. Mainly these are just missing a scheme, e.g. "www.kroger.com" instead of "http://www.kroger.com". This adds a pretty simple routine for cleaning those up (and also handling leading/trailing spaces, etc.). It's simple enough that I didn't think an external dependency was worthwhile, but if we wanted to do more, then we should probably use something pre-built.

Fixes https://sentry.io/organizations/usdr/issues/3106483109 and https://sentry.io/organizations/usdr/issues/3106483083.

### DIFF
--- a/loader/src/utils.js
+++ b/loader/src/utils.js
@@ -542,7 +542,7 @@ module.exports = {
     let result = text?.trim();
     if (!result) return undefined;
 
-    const urlPattern = /^(https?:\/\/)?[^/\s]+\.[^/\s]{2,}(?:\/.*)?$/i;
+    const urlPattern = /^(https?:\/\/)?[^/\s]+\.[^/\s]{2,}(?:\/[^\s]*)?$/i;
     const urlParts = result.match(urlPattern);
     if (urlParts) {
       if (!urlParts[1]) {
--- a/loader/src/utils.js
+++ b/loader/src/utils.js
@@ -523,4 +523,34 @@ module.exports = {
 
     return undefined;
   },
+
+  /**
+   * Ensure a string is a complete URL. If it is already a valid URL, the input
+   * string is returned as-is. If it's missing a scheme (e.g. "www.kroger.com"),
+   * the returned value will have a scheme added. If null, undefined, or an
+   * empty string are passed in, `undefined` will be returned.
+   *
+   * If the input doesn't look like it's really a URL (e.g. "Not a URL!"), this
+   * will throw a `ParseError`.
+   *
+   * This is meant for fairly simple scenarios; more we should use an external
+   * dependency if our needs get much more complex.
+   * @param {string} text Something that should be a URL.
+   * @returns {string|undefined}
+   */
+  cleanUrl(text) {
+    let result = text?.trim();
+    if (!result) return undefined;
+
+    const urlPattern = /^(https?:\/\/)?[^/\s]+\.[^/\s]{2,}(?:\/.*)?$/i;
+    const urlParts = result.match(urlPattern);
+    if (urlParts) {
+      if (!urlParts[1]) {
+        result = `http://${result}`;
+      }
+      return result;
+    }
+
+    throw new ParseError(`Text is not a URL: "${text}"`);
+  },
 };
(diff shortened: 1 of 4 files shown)


loader/test/utils.test.js
```diff
@@ -188,3 +189,32 @@ describe("parseUsPhoneNumber", () => {
...
+
+  it("returns undefined for empty input values", () => {
+    expect(cleanUrl(null)).toBe(undefined);
+    expect(cleanUrl("")).toBe(undefined);
+    expect(cleanUrl("    ")).toBe(undefined);
+  });
+
+  it("throws for values that are not URL-like", () => {
+    expect(() => cleanUrl("not a URL")).toThrow(ParseError);
+    expect(() => cleanUrl("http://no-second-level-domain")).toThrow(ParseError);
```
astonm, line 218:
```suggestion
    expect(() => cleanUrl("http://no-top-level-domain")).toThrow(ParseError);
```

loader/src/utils.js
```diff
@@ -523,4 +523,34 @@ module.exports = {
...
+   * This is meant for fairly simple scenarios; more we should use an external
+   * dependency if our needs get much more complex.
+   * @param {string} text Something that should be a URL.
+   * @returns {string|undefined}
+   */
+  cleanUrl(text) {
+    let result = text?.trim();
+    if (!result) return undefined;
+
+    const urlPattern = /^(https?:\/\/)?[^/\s]+\.[^/\s]{2,}(?:\/.*)?$/i;
```
astonm, line 545: Nitpicking, and I know it's supposed to be simple, but it looks like this accepts a URL with a space in it so long as it comes after the first `/` (e.g. `"http://thisshouldbe.com/invalid right?"` matches).

astonm (approved):
Looks pretty good and I imagine will catch the most egregious cases. Couple notes below for you for small tweaks.

Btw, another approach without pulling in a library but getting the benefits of a more rigorous parser would be to try `new URL(result)` and then if that fails also `new URL("http://" + result)` to see if appending the protocol fixes things.

loader/test/utils.test.js
Mr0grog: I beg to differ! The TLD here is `no-second-level-domain`! 😉

loader/src/utils.js
Mr0grog: Oh, good catch.

Mr0grog:
Quick thought on using the `URL` constructor… it ultimately results in more code (but I guess more detailed errors), since the regex here is checking for a lot more than pure *validity*. I started to implement your suggestion and then realized how it’s more complex in some ways (but maybe less obscure because there’s not as much regex):

```js
  cleanUrl(text) {
    let result = text?.trim();
    if (!result) return undefined;

    // The URL constructor will handle spaces loosely, escaping them as %20,
    // so we have to check for those ourselves.
    if (/\s/.test(result)) {
      throw Object.assign(new TypeError("URL should not have spaces"), {
        code: "ERR_INVALID_URL",
        input: text,
      });
    }

    // This might ultimately just be simpler if we just checked that it starts with `http://` or `https://`
    // or at least checked that it starts with a protocol (/^[A-Za-z][A-Za-z+\-.]*:\/\//) instead of parsing twice
    let parsed;
    try {
      parsed = new URL(result);
    } catch (_error) {
      result = `http://${result}`;
      parsed = new URL(result);
    }

    const parsed = new URL(result);
    if (parsed.protocol !== "http:" && parsed.protocol !== "https:") {
      throw Object.assign(new TypeError("URL must be HTTP or HTTPS"), {
        code: "ERR_INVALID_URL",
        input: text,
      });
    }
    if (!parsed.host.includes(".")) {
      throw Object.assign(
        new TypeError("URL did not have a second-level domain"),
        { code: "ERR_INVALID_URL", input: text }
      );
    }

    return result;
  },
```

It sounded good to me at first, but not sure it’s worthwhile. Although maybe I should still change `ParseError` to `TypeError` so it matches up with the URL constructor.

astonm:
Your change to disallow the space in the path looks good to me. I should have checked my own example to learn it would turn into a `%20`, heh.

Definitely a balance here between lots of code and lots of specific error messages vs. something short and sweet that accepts or rejects and moves on. I don't have a strong preference towards using `new URL` so 👍🏽 to just rolling with the regular expression code as it stands. Thanks for working through it fully.

PR #633, opened 2022-04-16
Add incident report for today's issue with updates

Mr0grog:
Pretty much just what it says on the tin!


astonm (approved):
lgtm!

Thanks for the quick and thorough write-up.

Issue #454, opened 2021-11-10
Set Nock-Back not to make new recordings in CI


Mr0grog:
The issue in #453 was masked by the fact that we always put Nock-Back in “record” mode during a test that uses it: https://github.com/usdigitalresponse/univaf/blob/4fc040bcb78c3397d27b9b85385b5c24dd2c22af/loader/test/support/nock-back-plugin.js#L67-L75

Instead, we should probably put it in “lockdown” mode in CI: https://www.npmjs.com/package/nock#user-content-modes

That probably means supporting an environment variable for this, or keying off an existing one this is set in CI environments like GitHub Actions, e.g. the `CI` variable. (Reference for actions: https://docs.github.com/en/actions/learn-github-actions/environment-variables)

PR #637, opened 2022-04-18
Fail CI if Nock.back recordings are missing

Mr0grog:
This makes sure Nock cannot connect to live servers and make new recordings during tests in CI, effectively ensuring that all HTTP requests in tests either:

1. Have saved recordings to use,
2. Have explicit mocks in code, or
3. Are explicitly allowed by test code.

Fixes https://github.com/usdigitalresponse/univaf/issues/454.


Mr0grog:
And this failed exactly as intended! 🎉

PR #654, opened 2022-05-13
Handle Bartell stores showing up in Rite Aid APIs

Mr0grog:
In hindsight, I probably should have seen this coming. Rite Aid bought Bartell back in 2020, and it looks like they may be starting to integrate systems — Bartell data is showing up in Rite Aid’s booking website backend, causing our scraper to break. See Sentry issue: https://sentry.io/organizations/usdr/issues/3141140661.

Unfortunately, we already have most Bartells via the CDC’s data portal, and we have them stored with a Bartell-specific ID rather than a Rite Aid-wide ID.

However! We happen to know that Rite Aid’s IDs are in the form `"<brand-based-prefix><brand-based-ID>"`, so we can use this appropriately determine both an ID for both the `rite_aid` and `bartell` external ID systems, and make sure all the right records get merged together. 🎉

The code on the CDC side is a little messy, since it originally supported only one external ID per location, and now we need to support two. I could have done a large refactoring, but opted not to for now.

This will probably need to be followed up with a run of the merge script to fix some locations we’ve previously been identifying as separate based on the Rite Aid API and the CDC open data. For example, these are the same:
- http://getmyvax.org/api/edge/locations/bartell:10
- http://getmyvax.org/api/edge/locations/rite_aid:6910

Fixes https://sentry.io/organizations/usdr/issues/3141140661

### DIFF
--- a/loader/src/sources/cdc/api.js
+++ b/loader/src/sources/cdc/api.js
@@ -347,7 +347,7 @@ function getStoreExternalIds(location) {
           ? idValue
           : [[definition.system, idValue]];
       }
-      return [];
+      break;
     }
   }
 
--- a/loader/src/sources/cdc/api.js
+++ b/loader/src/sources/cdc/api.js
@@ -94,8 +94,8 @@ function formatStore(storeItems, checkedAt) {
       provider: "cdc",
     });
 
-    const storeExternalId = getStoreExternalId(base);
-    if (!storeExternalId) {
+    const storeExternalIds = getStoreExternalIds(base);
+    if (!storeExternalIds.length) {
       return null;
     }
 
@@ -140,10 +140,10 @@ function formatStore(storeItems, checkedAt) {
     result = {
       external_ids: [
         ["vaccines_gov", base.provider_location_guid],
-        storeExternalId,
+        ...storeExternalIds,
       ],
       name: titleCase(base.loc_name),
-      provider: storeExternalId[0],
+      provider: storeExternalIds[0][0],
       address_lines: addressLines,
       city: titleCase(base.loc_admin_city),
       state: base.loc_admin_state,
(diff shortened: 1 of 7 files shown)


loader/src/sources/riteaid/common.js
```diff
@@ -9,4 +9,46 @@ class RiteAidApiError extends HttpApiError {
...
+  let riteAidId;
+  for (const [system, value] of externalIds) {
+    if (system === "rite_aid") {
+      riteAidId = value;
+    } else if (system === "bartell") {
+      return `Bartell Drugs #${value}`;
+    }
+  }
+
+  return `Rite Aid #${riteAidId}`;
```
astonm, line 51: Maybe make returning this conditional on whether `riteAidId` got set to a useful value?

Alternately, could return immediately on finding a `rideAidId` value.

loader/src/sources/cdc/api.js
```diff
@@ -324,18 +333,25 @@ const mislabeledLocations = {
...
   for (const definition of locationSystems) {
     if (definition.pattern.test(location.loc_name)) {
       const idValue = (definition.getId || getSimpleId)(location);
-      return idValue ? [definition.system, idValue] : null;
+      if (idValue) {
+        return Array.isArray(idValue)
+          ? idValue
+          : [[definition.system, idValue]];
+      }
+      return [];
```
astonm, line 350: Do you need this line, or can you safely fall through to the similar `return []` at the end of the function?

astonm (approved):
Cool, lgtm. Couple small notes:

loader/src/sources/riteaid/common.js
Mr0grog: Ah, yeah, fair point. This is outside the loop because I was trying to be sensitive to the fact that the Rite Aid ID and sub-brand ID could come in any order, and we want to prefer the sub-brand. (In practical terms, this is always coming from the `getExternalIds()` function, so we could just iterate in reverse order, and then the first one would always be the more specific one.

loader/src/sources/cdc/api.js
Mr0grog: Ugh, the revision here probably made things too subtle: it can’t fall through because it needs to stop iterating once it’s found a matching entry in `locationSystems`. This matches the existing behavior, where an empty result from the location system causes the location to be dropped from the output. (OTOH, I can see a version of the universe where that should be an error we get alerted about, but… that seems like a big change to make here.)

loader/src/sources/riteaid/common.js
Mr0grog: Rejiggered this a bit to cover not finding a `rite_aid` system, and maybe read a little bit more clearly about why it doesn’t return immediately.

loader/src/sources/cdc/api.js
Mr0grog: Realized I could switch this to a `break`, which is a *little* clearer, and means we don’t repeat `return []` in multiple places.

astonm (approved):
New changes help with clarity, thanks!

PR #673, opened 2022-05-24
Automatically deploy in CI workflow on main branch

Mr0grog:
Manually doing multiple steps, and waiting for the right timing is both frustrating and error-prone. I should have long since just automated deployment from `main`.


astonm (approved):
Assuming `./scripts/deploy_infra.sh` does the right thing, 👍🏽

Mr0grog:
I hope so, since you wrote it! 😉

PR #676, opened 2022-05-25
Clean up messy new names for Albertsons pediatric booster locations

Mr0grog:
This is *way* not ideal, but there are now a whole mess of locations from Albertsons without store numbers in the address. The best I could think to do without massive change here is a bunch of overrides, but I have a feeling we’re going to see more of these, and will have to come up with something fancier.


Mr0grog:
Fixes https://sentry.io/organizations/usdr/issues/2829716008

Mr0grog:
Going ahead and merging this to clean up things that are actively breaking.

PR #677, opened 2022-05-25
Add scraper to find every Albertsons pharmacy

Mr0grog:
This is a follow-on to #676 and is a bit speculative at the moment. We’re seeing an increasing number of locations with no store number, but we *can* scrape a list of [deeply] detailed info for every Albertsons pharmacy (would you like to know if a store has fresh guacamole? Now you can find out!).

This is sort of a proof-of-concept; I’m thinking we’d only run this very rarely and manually for now and commit the resulting data files (but slimmed down; these are HUGE) so we can do lookups either on addresses or latitude/longitude. It takes a while and a lot of resources to do this scrape, and the parts we care about should chaange exceedingly rarely.


Mr0grog:
Well, I’ve now opened a whole new can of worms:
- The data I scraped here has a field named `c_oldStoreID` which is often different from `c_parentEntityID` and `c_associatedPharmacyStoreID`. I’m guessing these might be store numbers a sub-brand used before they got bought by Albertsons.
- Sometimes the Albertsons *appointment* data uses the *old* store number from above instead of a current/new one.

So this is now repairable in a lot of cases, but will be a real serious mess. If we want to do this cleanly, we should probably:
- Update the Albertsons loader to use external ID systems like `albertsons_old_xxx` or something similar instead of `albertsons_xxx`.
- Add a migration to Update all the existing external ID records in the database in a similar way.
- When we land this stuff, switch the scraper external ID systems back to `albertsons_xxx`

Less work, but also less nice, would simply be using different external IDs when we land this, e.g. `albertsons_v2_xxx` instead of `albertsons_xxx`.

Mr0grog:
OTOH, matching with this data finds stores for two places that are currently getting classified as community clinics because they have no store name in the appointment data. 🙃

Mr0grog:
This is also finding a handful of situations where the address in the appointment data is *just wrong,* e.g:
- Appointments: [Acme 3970 - 631 Mantua Ave, West Deptford, NJ, 08096](https://www.google.com/maps/place/631+Mantua+Ave,+West+Deptford,+NJ+08096/@39.8375449,-75.2416881,17z/data=!3m1!4b1!4m5!3m4!1s0x89c6db5a63641eb3:0x9ca161d2dafb892c!8m2!3d39.8375449!4d-75.2394994) (This is a real address, but is a house 6.98 km away from the store in question.)
- Store data: [ACME #3970 - 631 Mantua Pike, Woodbury, NJ 08096](https://www.google.com/maps/place/631+Mantua+Pike,+Woodbury,+NJ+08096/@39.8270559,-75.1634759,17z/data=!3m1!4b1!4m5!3m4!1s0x89c6daa5e96bcbd9:0x755876f828f55656!8m2!3d39.8270559!4d-75.1612872)

Mr0grog:
Rebased and added some terrifying in-progress/experimental code, but just wanted to make sure it was saved. This seems to get a pretty good match in all cases, finds more correct store numbers or addresses when the appointment data may be wrong (at least according to way too many spot checks), and successfully finds brand info when it is entirely missing and/or totally unmatchable in the appointment data.

Mr0grog:
Quick update: I’ve been a bit busy with offboarding tasks related to leaving USDR, but expect to get back to this over the weekend or next week.

Mr0grog:
I also tested this against the CDC API and it looks like they universally use the “new” store numbers. (Bad news is that means we are probably already mixing things up for locations that had different old & new numbers.)

Mr0grog:
In terms of different old vs. new store numbers:
- There are 715 stores where old and new numbers are different.
- None are in Alaska
- 58 are in New Jersey. These cover two brands:
    - ACME Markets: 41
    - Kings Food Markets: 17
- Over *all* stores, most of the brands have some flavor of this problem:
    - Albertsons: 257
    - Shaw's: 127
    - ACME Markets: 96
    - Jewel-Osco: 76
    - United Supermarkets: 48
    - Albertsons Market: 23
    - Star Market: 20
    - Kings Food Markets: 19
    - Market Street: 19
    - Safeway: 15
    - Balducci's: 8
    - Amigos: 4
    - Lucky: 1
    - Tom Thumb: 1
    - Randalls - CLOSED: 1

Mr0grog:
Feeling pretty good about how this works now. (Obviously I need to clean out all the crazy logging code, make it *use* the new data, and fix tests, though.)

Here’s a listing of all the locations that will wind up with different store numbers after matching: https://gist.github.com/Mr0grog/119c526c73b5c7e122640e99dae94d4f

There are two locations that will need special handling:

```
Method: fuzzy  / Score: 0.2  / Factors: [ 'geo' ]
Appointments: Vons #1738                               /  729 North H St, Lompoc, CA 93436  /  VONS #1738  - 729 North H St, Lompoc, CA, 93436
Match:        Albertsons Pharmacy #3171                /  1500 N H St, Lompoc, CA 93436
              (c_oldStoreID: 3171)
```

This Vons is not listed on the Vons website, but appears to maybe be real? If we only take scores > 0.2 (which seems plenty reasonable) that’ll solve this.

```
Method: fuzzy  / Score: 0.55  / Factors: [ 'brand', 'number', 'geo' ]
Appointments: United Supermarkets #573                 /  101 N. Washington, Seymour, TX 76380  /  United 0573 - 101 N. Washington, Seymour, TX, 76380
Match:        United Supermarkets Pharmacy #146        /  202 N Main St, Seymour, TX 76380
              (c_oldStoreID: 573)
```

Google Maps suggests these really are two different locations, even though the pharmacy in question ([101 N. Washington in Seymour, TX](https://www.google.com/maps/place/101+N+Washington+St,+Seymour,+TX+76380/@33.5881333,-99.2641639,17z/data=!3m1!4b1!4m5!3m4!1s0x86548374024e5009:0x4f85c49077cb8403!8m2!3d33.5881289!4d-99.2619752), as opposed to [the one we're matching to](https://www.google.com/maps/place/202+N+Main+St,+Seymour,+TX+76380/@33.5895391,-99.261952,17z/data=!3m1!4b1!4m5!3m4!1s0x865483739a9b4a6f:0xca602d27ff42d94f!8m2!3d33.5895347!4d-99.2597633)) is not listed on United Supermarkets’s (or any Albertsons sub-brand’s) website. Street view shows a small pharmacy-only outlet at this address.

I think the best solution for this (and maybe something we should do for the Vons just in case) is to list them in the `corrections.js` module with some extra key like `doNotMatchKnownStores: true` or something that tells the rest of the code not to try and look up that store. These are both just weird cases where Albertsons lists nothing about them publicly, but that appear to potentially be real stores. The matching code is doing a great job finding other stores of the same brand less than a 1km away that would normally seem like the correct listing.

Mr0grog:
GOOD NEWS! I was worried about the changing store numbers causing collisions, but it looks like there are somehow no instances of this in practice!

Full details on how I did this and every change and collision are in this gist: https://gist.github.com/Mr0grog/0f832b801bdf52896d30f8656c54de98

But here’s the quick summary:
- **147** locations are getting a different (Store Brand, Store Number) combination.
- **34** of these collide with an existing location listing in UNIVAF’s database.
- **12** of these are potentially concerning because they are changing the store number (as opposed to adding a number to a listing we simply weren't able to detect one in before) or they are changing the brand from another *store* brand, e.g. Pavilions → Safeway (as opposed to from a “community clinic” to a store).
- **0** of the collisions appear to represent a different place -- they all have the same address as the corrected appointment listing we want to write.

So it looks like we don’t actually have to worry about this before merging.

On the other hand we should come back after merging and delete the old external IDs for every one of the 147 changed locations that are changing the store number (as opposed to adding a number to a listing we simply weren't able to detect one in before) or are changing the brand from another *store* brand, e.g. Pavilions → Safeway (as opposed to from `community_clinic`).

🎉

loader/src/sources/albertsons/index.js
```diff
@@ -480,6 +488,82 @@ function formatLocation(data, validAt, checkedAt) {
...
+      !["albertsons_corporate", "community_clinic"].includes(storeBrand?.key)
+    ) {
+      console.error("NO MATCH:", data.address);
+      const appAddress = `${address.lines[0]}, ${address.city}, ${address.state} ${address.zip}`;
+      console.error(
+        `  Parsed: ${storeBrand?.name} #${storeNumber}, ${appAddress}`
+      );
+      console.error("------------");
+    }
+  }
```
astonm, line 540: Mind moving this out to its own function?

loader/src/sources/albertsons/scrape-albertsons-stores.js
```diff
@@ -0,0 +1,305 @@
...
+ * @yields {any}
+ */
+async function* crawlListingPage(pageUrl, depth = 0, startIndexes = []) {
+  if (depth > 4) throw new Error(`Too many listing pages deep: "${pageUrl}"`);
+
+  let urls = await withPage({ retries: 1 }, async (page) => {
+    await navigate(page, pageUrl);
+
+    // If the state only has one store, this will actually be a store page.
+    const locationData = await getLocationDataFromStorePage(page);
```
astonm, line 233: Incomplete refactor here?

astonm (approved):
Wow, awesome that it all worked out in the end. Appreciate all of your work ensuring we got everything right here, and code looks good to me (modulo a couple tiny things).

loader/src/sources/albertsons/index.js
Mr0grog: Ah yeah, I was debating between that and just removing it entirely (I was mainly keeping it in until things were ready to merge in case I needed to re-do any of the above analyses). Will make sure to do one of those.

loader/src/sources/albertsons/scrape-albertsons-stores.js
Mr0grog: Heh, yep, I messed up that commit. Will fix.

loader/src/sources/albertsons/scrape-albertsons-stores.js
```diff
@@ -20,6 +20,59 @@
...
+ *         c_parentEntityID,
+ *         c_pharmacyBrand,
+ *         geocodedCoordinate,
+ *         name,
+ *         timezone,
+ *         covid19InformationUrl,
+ *         covidVaccineAppointmentRequired,
+ *         covidVaccineSiteInstructions,
+ *         description,
+ *        }]' raw-output.json > albertsons-pharmacies.json
```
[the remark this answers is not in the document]
Mr0grog: Fixed in 788916eebe434295f0941c6e87f873c2170b9b05

loader/src/sources/albertsons/scrape-albertsons-stores.js
Mr0grog: Should be fixed in 788916eebe434295f0941c6e87f873c2170b9b05

loader/src/sources/albertsons/index.js
Mr0grog: Kept it, and moved to a separate function.

Mr0grog:
Assuming this passes tests, it *should* be good to go. @astonm thank you so much for taking a look earlier, and feel free to review the last few changes I made (mostly on your feedback). I’m planning to merge in the late afternoon/early evening.

astonm:
Cool, changes all look good to me! 🚢

Mr0grog:
Update: all the follow-on stuff is done now. There was one correction I had to add (a badly geocoded location was causing a place to match to the wrong saved store data), 9 weird entries that had collided (I think they may have been colliding even before this change), and then all the old external IDs to remove.

In the case of old external ID removals, they are all still available (in case we discover we did something bad) in `meta.former_external_ids`. For example, in https://getmyvax.org/api/edge/locations/be2c9f41-0594-40cf-97ba-51c052cc6746

```js
{
  id: "be2c9f41-0594-40cf-97ba-51c052cc6746",
  provider: "albertsons_united",
  location_type: "PHARMACY",
  name: "United 0533 United",
  address_lines: ["3400 River Road"],
  city: "Amarillo",
  state: "TX",
  postal_code: "79107",
  county: null,
  position: {
    longitude: -101.8322757,
    latitude: 35.2481958,
  },
  info_phone: null,
  info_url: null,
  booking_phone: null,
  booking_url: "https://www.albertsons.com/pharmacy/covid-19.html",
  description: null,
  requires_waitlist: false,
  meta: {
    time_zone: "America/Chicago",
    vaccinespotter: {
      brand: "united",
      brand_id: 18,
      provider: "albertsons",
    },
    former_external_ids: [
      ["albertsons_store_number", "united:0533"],
      ["albertsons_store_number", "united:533"],
      ["united", "0533"],
      ["united", "533"],
    ],
  },
  created_at: "2021-05-24T17:23:48.240Z",
  updated_at: "2021-05-24T17:23:48.240Z",
  external_ids: [
    ["albertsons", "1600104192608"],
    ["albertsons_store_number", "united:133"],
    ["albertsons_united", "1600104192608"],
    ["united", "133"],
    ["vaccinespotter", "1046423"],
  ],
  availability: {
    sources: ["univaf-albertsons"],
    valid_at: "2022-06-10T22:16:19.000Z",
    checked_at: "2022-06-10T22:27:07.716Z",
    available: "NO",
    changed_at: "2021-11-28T20:38:28.000Z",
  },
}
```

PR #696, opened 2022-06-10
Handle new Walmart store name formatting from CDC

Mr0grog:
When Walmart stores had a store number in the name (instead of in the store number field), they used to be formatted as "Walmart NNN" (where N is a digit). Some are now appearing with a "10-" prefix, e.g. "Walmart 10-0178" (the same format as in the store number field). This alters our parser to handle both.

Fixes https://sentry.io/organizations/usdr/issues/3338103303.

I also fixed a geographic position validation bug I noticed while I was addressing the above issue.

### DIFF
--- a/loader/src/sources/cdc/api.js
+++ b/loader/src/sources/cdc/api.js
@@ -202,10 +202,10 @@ function getWalmartId(location) {
     return unpadNumber(location.loc_store_no.slice(3));
   }
 
-  // When the number is not in loc_store_no, it is usually formatted as a normal
-  // store number in the name (i.e. "NNN" instead of "10-NNN").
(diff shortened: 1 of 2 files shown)


loader/src/sources/cdc/api.js
```diff
@@ -166,7 +166,7 @@ function formatStore(storeItems, checkedAt) {
         longitude: Number(base.longitude),
         latitude: Number(base.latitude),
       };
-      if (isNaN(position.longitude) || isNaN(position.longitude)) {
+      if (isNaN(position.longitude) || isNaN(position.latitude)) {
```
astonm, line 169: ❗

astonm (approved):
👍🏽

PR #704, opened 2022-06-15
Test whether Terraform state is good

Mr0grog:
This is a test of what Terraform Cloud is currently trying to do. #703 failed to deploy because the DNS record it added already exists, which seems reasonable. This is a test to see whether things are all good now, or if it will be stuck until I delete that record and let Terraform recreate it.


Mr0grog:
NOPE. Terraform is still trying to recreate the `www.` DNS record. Going to delete it and let TF do the job.

Mr0grog:
Done remotely on AWS console and TF Cloud.

PR #706, opened 2022-06-15
HOTFIX: Disable primary host redirects

Mr0grog:
The primary host redirects we just added worked great in my testing, but do not seem to be playing nice with CloudFront and/or ELB (or maybe it’s because we have two levels of proxying there…?). This disables the redirects and adds logging in their place to see what’s going on.


Mr0grog:
See also:
- https://sentry.io/organizations/usdr/issues/3350105535
- https://sentry.io/organizations/usdr/issues/3350112992

PR #708, opened 2022-06-15
Forward the Host header from Cloudfront

Mr0grog:
Part of the problem in #706 is that the `Host` header is not getting forwarded by Cloudfront. This should solve that, although I suspect there might be some ELB rejiggering needed afterwards.

### DIFF
--- a/terraform/api.tf
+++ b/terraform/api.tf
@@ -214,6 +214,7 @@ resource "aws_cloudfront_distribution" "univaf_api" {
     max_ttl                = 3600
 
     forwarded_values {
+      headers = ["Host", "Origin"]
       query_string = true
 
       cookies {


terraform/api.tf
```diff
@@ -214,6 +214,7 @@ resource "aws_cloudfront_distribution" "univaf_api" {
     max_ttl                = 3600
 
     forwarded_values {
+      headers = ["Host", "Origin"]
```
Mr0grog, line 217: If I understand the docs right, this will both forward these headers and include them in the cache key. I think that’s reasonable for these, but if we have others we only want to forward in the future (e.g. `User-Agent`), I think we have to create separate [`aws_cloudfront_origin_request_policy`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_origin_request_policy) and [`aws_cloudfront_cache_policy`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_cache_policy) resources.

Issue #302, opened 2021-07-12
Serve API at www.getmyvax.org


astonm:
Currently the API lives at https://getmyvax.org but does not resolve if you visit https://www.getmyvax.org/ instead.

We should support both URLs, either separately or by forwarding from one to the other is fine.

Mr0grog:
Unfortuantely not solved. This turned into a bit of a mess (see #705 and #706) and I had to revert some of the PR that was supposed to resolve this issue.

In #706 I added some logging, and it looks like the main problem here is that headers aren’t getting forwarded correctly. Here’s an example log:

```
2022-06-15T05:17:40.810Z [univaf] info: DEBUG HOST for GET /api/edge/locations?state=NY&cache=16
primaryHost: "getmyvax.org"
getRequestHost: "api-load-balancer-830424586.us-west-2.elb.amazonaws.com"
Host header: "api-load-balancer-830424586.us-west-2.elb.amazonaws.com"
X-Forwarded-Host header: "undefined"
request.socket.remoteAddress: "::ffff:172.17.3.233"
trust: true
```

Mr0grog:
#708 looks to have solved the underlying issue; I am now seeing logs like:

```
2022-06-15T06:18:29.679Z [univaf] info: DEBUG HOST for GET /api/edge/locations?state=OR&cache=20
primaryHost: "getmyvax.org"
getRequestHost: "www.getmyvax.org"
Host header: "www.getmyvax.org"
X-Forwarded-Host header: "undefined"
request.socket.remoteAddress: "::ffff:172.17.2.253"
trust: true
```

It’s a little late to mess with deploying new JS code tonight, but it looks like this *should* be ready to flip the redirects back on tomorrow morning (i.e. revert #706).

---

On the other hand… it looks like there are some ways to do this purely in either CloudFront or ELB that I missed before:

- CloudFront functions are JS snippets you can run on the edge in CloudFront. Here’s a blog post about redirects: https://tech.mybuilder.com/handling-redirects-with-cloudfront-functions/ and some official docs: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/example-function-redirect-url.html

- ELBs can have listener rules — [`lb_listener_rule`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb_listener_rule) in Terraform — that do redirects and can be conditioned based on the Host (or any other) header.

[...]

PR #709, opened 2022-06-15
Implement `www.` → no subdomain redirect in ELB

Mr0grog:
I thought I'd looked for a way to do this before and didn't see it, but while dealing with issues in the earlier approach, I found this. Hopefully this (plus #708, which this depends on) solves #302 in pretty much the ideal way.

If this works out, there’ll be some cleanup to do from the earlier approach.


Mr0grog:
HERE WE GO

PR #711, opened 2022-06-15
Cleanup after `www` redirect mess

Mr0grog:
Well, #665 turned into quite a mess yesterday! First it had some problems with Terraform (#703, #704), then healthcheck issues that caused ECS to retract the deployment (#705), and then the redirects didn’t actually work in production and caused an infinite loop, so I temporarily disabled them and replaced them with logging (#706). The logs showed that the app wasn’t receiving the expected `Host` header or an `X-Forwarded-Host` proxy header, which we solved by fixing the Cloudfront configuration (#708). Finally, I realized there was a way to do the redirects at the ELB level instead of the app (I’d looked for this originally and apparently missed it), so I went ahead and did that (#709).

Now things are working great! 🎉

However, all that left a little detritus along the way, and this PR is a quick cleanup attempt.


astonm (approved):
🎉 

Nice solution, in the end. And we learned so much along the way 😄

Issue #710, opened 2022-06-15
Support for 6 month + vaccine products


dhcole:
We're hoping to see approval for new vaccine products, "Modern 6 month - 5 year", and a "Pfizer 6 month - 4 year". Is the a code we should look for for these coming from univaf sites? I believe we have an unused `pfizer_age_2_4` that may be repurposed.

Mr0grog:
We can also rename this to `pfizer_age_0_4` real quick if that would be better.

Otherwise, we’re on it! For sources that use CVX codes (some SMART Scheduling Links sources, but not all) or NDC codes (the CDC), this should just work. For all the others (e.g. Albertsons chains, Rite Aid), we’ve got alerts for whenever we encounter a new vaccine name that should help us get these hooked up as soon as they start showing up in source data. 👍

Mr0grog:
Will also add a `moderna_age_0_5` for the corresponding Moderna vaccine, which it looks like was also approved.

Mr0grog:
Just to clarify, @dhcole, I’d like to rename it to `pfizer_age_0_4`, but want to make sure you aren’t already expecting/depending on the old name first. So let me know if that’s ok, please!

dhcole:
Yes renaming is fine with us. Thanks @Mr0grog

Mr0grog:
Alright, I renamed `pfizer_age_2_4` to `pfizer_age_0_4` and did a bunch of other cleanup and screw-tightening on vaccine types over in #714.

So the vaccine products you should expect to see in UNIVAF in the future are:
- `astra_zeneca`
- `jj`
- `moderna`
- `moderna_age_6_11`
- `moderna_age_0_5`
- `novavax`
- `pfizer`
- `pfizer_age_5_11`
- `pfizer_age_0_4`
- `sanofi` (not actually approved for use in the US yet)

I think we’ve done everything that can reasonably done in advance here, so I’m going to close this issue. Thanks for raising it, @dhcole!

PR #714, opened 2022-06-15
Add constants and codes for early childhood

Mr0grog:
Vaccines for ages 6 months - 4 years (Pfizer) and 6 months - 5 years (Moderna) were approved today, and this adds constants for them as well as support for their known NDC and CVX codes so we are likely to be able to recognize them in the data.

Partially covers #710.


loader/src/smart-scheduling-links.js
```diff
@@ -38,11 +38,16 @@ const EXTENSIONS = Object.freeze({
 const PRODUCTS_BY_CVX_CODE = {
   210: "astra_zeneca",
   207: "moderna",
+  221: "moderna",
   211: "novavax",
   208: "pfizer",
+  217: "pfizer",
   218: "pfizer_age_5_11",
   219: "pfizer_age_2_4",
```
astonm, line 46: Any reason not to go ahead and rename this `pfizer_age_0_4`?

astonm (approved):
👍🏽

loader/src/smart-scheduling-links.js
Mr0grog: Yep — NJ already knows about and is using it. I’d like to rename, but want a go-ahead from them first: #710.

Mr0grog: Got a 👍 there so will also tweak that name before merging.

Mr0grog:
OK! I renamed `pfizer_age_2_4` to `pfizer_age_0_4` and did a bunch of cleanup for all our fuzzy name matching to make it a little bit more strict and alert us when we see new kinds of pediatric vaccines. I also made it match the new 6 months - 4/5 years names that are currently showing up in Alaska’s PrepMod staging site (nothing in production for them yet).

Since I had to mess with the fuzzy matching, I also ran all the loaders to make sure they aren’t going to fail matching a vaccine that is currently working in production today.

astonm (approved):
Cool, lgtm. :shipit:

Mr0grog:
Thanks!

PR #724, opened 2022-06-22
Increase POST body limit to 2.5 MB

Mr0grog:
We are now seeing a handful of GIANT listings from PrepMod which appear to be legit, so this updates the allowed body size to handle them.

For example, a clinic in Puyallup, WA has 3897 slots (!) from 47 separate events (looks like a morning & afternoon event each Wednesday - Saturday) for the next 6 weeks. They’re really scheduling ahead (probably because of the new early pediatric vaccines).


Mr0grog:
Fixes https://sentry.io/organizations/usdr/issues/3233514222

PR #726, opened 2022-06-22
Use the standard loader logger everywhere

Mr0grog:
**This is a follow-on to/depends on #725.** Since I was updating logging, it seemed like it made sense to normalize everything. But I hadn’t realized how many loaders we never updated, and this turned into a *way* bigger job than expected. 😬

There is probably some good refactoring to be done on the logging tools, but that'll be easier after making everything use the tools we currently have.

This also sends a bit more context data to Sentry. I’m thinking we should just always do that instead of making it optional (I think I may have originally been under the impression that Sentry couldn’t take complex objects in the context, but [this test shows it actually works great](https://sentry.io/organizations/usdr/issues/3368356106/)), but that’s a refactor that should probably be reserved for a separate PR.

### DIFF
--- a/loader/test/albertsons.test.js
+++ b/loader/test/albertsons.test.js
@@ -10,6 +10,9 @@ const { expectDatetimeString, splitHostAndPath } = require("./support");
 const { locationSchema } = require("./support/schemas");
 const { ParseError } = require("../src/exceptions");
 
+// Mock utils so we can track logs.
+jest.mock("../src/utils");
+
 const [API_URL_BASE, API_URL_PATH] = splitHostAndPath(API_URL);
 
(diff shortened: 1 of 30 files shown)


loader/test/albertsons.test.js
```diff
@@ -10,6 +10,9 @@ const { expectDatetimeString, splitHostAndPath } = require("./support");
 const { locationSchema } = require("./support/schemas");
 const { ParseError } = require("../src/exceptions");
 
+// Mock utils so we can track logs.
+jest.mock("../src/utils");
+
```
Mr0grog, line 15: Tests are SO MUCH more readable now. 😅

PR #733, opened 2022-06-23
Add minimum age support for locations

### LINKED ISSUE #732: Accessing minimum age metadata
Pharmacies may receive a vaccine product for young children (6mo+) but only be authorized to administer it to individuals 3 year+. Vaccines.gov seems to have this metadata represented by a `minimum_age` attribute: https://api.us.castlighthealth.com/vaccine-finder/v1/provider-locations/97650e43-14be-4275-bcaf-ca28aceba3da?inventoryCategory=covid

Having access to this attribute in the univaf feed would enable us to show users which sites are actually able to serve 6m-2y, vs which can only serve 3y+.

Would it be possible to include a `minimum_age` attribute on providers, where available?

Mr0grog:
This adds two fields to location: `minimum_age_months` and `minimum_age_years` to the schemas, API server, and database. It also *sources* that data from the CDC loader.

This is a first pass and we might tweak the format slightly.

Fixes #732.


Mr0grog:
This is nice and relatively simple, but I’m leaning a *little* bit towards changing it to an object, which might be nicer to document and use (and makes for slightly smaller JSON):

```js
{
  id: "abc-123"
  // ...other properties...
  minimum_age: {
    months: null,
    years: 12
  }
}
```

Mr0grog:
The other question here is whether it would be better to ensure it’s always a number (so we’d put `0` if unknown or if you should look at the other field, rather than `null` for this case).

astonm (commented):
I'm not sure if there's a good way to change the naming to make it obvious, but I wasn't sure until I read the docs whether the fields were mutually exclusive or whether they were additive (like "I'm 5 years and 3 months old"). Given that it's the former, it might be better to store a single value `min_age_months` and then if given a value in years, multiply by 12?

astonm:
Yeah, I think keep the raw data just in case. What sorts of conflicts did you see?

Mr0grog:
> What sorts of conflicts did you see?

Typically a `min_age_months` of 3 or 6 (sometimes 1!) and a `min_age_years` of 12 or 18. Looking again, though, there are only 15 locations with conflicts. I guess they cleaned up a lot! This view should show just the conflicts: https://data.cdc.gov/Vaccinations/Vaccines-gov-COVID-19-vaccinating-provider-locatio/5jp2-pgaw/explore/query/SELECT%0A%20%20%60provider_location_guid%60%2C%0A%20%20%60min_age_months%60%2C%0A%20%20%60min_age_years%60%2C%0A%20%20%60loc_name%60%0AWHERE%20%28%60min_age_months%60%20%3E%20%270%27%29%20AND%20%28%60min_age_years%60%20%3E%20%270%27%29%0AGROUP%20BY%0A%20%20%60provider_location_guid%60%2C%0A%20%20%60min_age_months%60%2C%0A%20%20%60min_age_years%60%2C%0A%20%20%60loc_name%60/page/aggregate

Mr0grog:
OK, I think this is probably good to go. Two major changes:

- There’s only one field on the location: `minimum_age_months`. It simply states the age in months, whether that’s 6 (should be the earliest possible) or 144 (12 years). **It may be null** if we don’t have this info for a location yet.

    If needed, you can did into `location.meta.cdc_minimum_age` to get the raw, unparsed values from the CDC.

- As a more useful shortcut, I added `location.availability.minimum_eligible_age_months`, which calculates the actual eligible age by looking at all the available vaccines and the minimum age for the location. For example, if the location’s minimum is 6 but the only vaccine offered is `pfizer`, this will be 144. But if `pfizer_age_0_4` is offered, then it will be 6.

PR #734, opened 2022-06-24
Handle early childhood vaccines from Albertsons

Mr0grog:
This got pretty messy. In https://sentry.io/organizations/usdr/issues/3374158962, we started seeing a new vaccine code (`ModernaChild`) for the new Moderna vaccine for ages 6 months – 5 years. I started by adding that and discovered two new problems along the way:

1. A lot of locations have separate booking URLs for the new Moderna and Pediatric vaccines. We previously solved this issue for New Jersey by adding `meta.booking_url_adult` and `meta.booking_url_pediatric`, and I added a `meta.booking_url_infant`, but now we have a situation where there are *multiple* infant URLs and *multiple* pediatric URLs.

2. While updating the tests, I noticed some locations labeled as “Pfizer infant” even though no warnings were getting logged for a new vaccine code for Pfizer. It turns out that Albertsons is using `PfizerChild` as a code for *both* the early pediatric vaccine (6 months – 4 years) and the existing pediatric one (5–11 years). 😭

More details on solutions for each of these below. If anyone has time, I’d love some feedback on these.

## Multiple Booking URLs

For backwards compatibility, I added `meta.booking_url_infant`. So an Albertsons location’s `meta` used to look like:

```js
{
  "timezone": "America/Los_Angeles",
  "booking_url_adult": "https://kordinator.mhealthcoach.net/vcl/1600119864549",
  "booking_url_pediatric": "https://kordinator.mhealthcoach.net/vcl/1636005234623"
}
```

And it now looks like:

```js
{
  "timezone": "America/Los_Angeles",
  "booking_url_adult": "https://kordinator.mhealthcoach.net/vcl/1600119864549",
  "booking_url_pediatric": "https://kordinator.mhealthcoach.net/vcl/1636005234623",
  "booking_url_infant": "https://kordinator.mhealthcoach.net/vcl/1656025272707"
}
```

Except that’s not quite true. It’s now possible to have more booking URLs than that can represent, so I *also* added a new `booking_urls` property that is meant to replace the above `booking_url_*` properties. It’s an array where each entry is an object like:

```js
{
  url: "",                // the URL
  products: [],           // array of products that are bookable here
  available: "YES" | "NO" // whether there are booking appointments at this link
}
```

So in practice, you might have this `meta` object for a location:

```js
{
  "timezone": "America/Los_Angeles",

  // Deprecated but present for backwards-compatibility
  "booking_url_adult": "https://kordinator.mhealthcoach.net/vcl/1600119864549",
  "booking_url_pediatric": "https://kordinator.mhealthcoach.net/vcl/1636005234623",
  "booking_url_infant": "https://kordinator.mhealthcoach.net/vcl/1656025272707",

  // New booking URL list
  "booking_urls": [
    {
      "products": ["pfizer", "moderna"],
      "url": "https://kordinator.mhealthcoach.net/vcl/1600119864549",
      "available": "YES"
    },
    {
      "products": ["pfizer_age_0_4"],
      "url": "https://kordinator.mhealthcoach.net/vcl/1656025272707",
      "available": "NO"
    },
    {
      "products": ["moderna_age_0_5"],
      "url": "https://kordinator.mhealthcoach.net/vcl/1656025553071",
      "available": "NO"
    },
    {
      "products": ["pfizer_age_5_11"],
      "url": "https://kordinator.mhealthcoach.net/vcl/1636005234623",
      "available": "YES"
    }
  ]
}
```

If this works for our partners, my plan is to get rid of `meta.booking_url_*` once they transition to using `meta.booking_urls`. I’d also like to do this for PrepMod if it’s good, since that’s another source where we have a multiple booking URLs issue.

Caveat: it’s not great that we are stuffing availability info into the location listing now. Some alternatives:
- Put the same info in `availability.meta.booking_urls` (`availability.meta` has always been there, but we’ve never used it).
- If we think this is really solid, we could make it a top-level field on availability: `availability.booking_urls`. This is a lot more work, though.

## `PfizerChild` has multiple meanings

Albertsons is using `PfizerChild` to represent both vaccines for 6 months – 4 years and 5 – 11 years. Right now, the code in this PR looks at the locations name (which comes from the `address` field in the source data) to see if it says the location is for infants. If so, it treats `PfizerChild` as indicating our `pfizer_age_0_4` code and if not, as our `pfizer_age_5_11` code.

It’s not clear that this is correct, though…

- For most of the “infant” listings, going to the booking URL in the listing lets you choose either pediatric type: Screen Shot 2022-06-23 at 4 46 03 PM

- For *some* of the non-infant listings, going to the booking URL in the listing presents you with an interstitial that only discusses “children aged 5 to 11” and not ages 3–4. But if you click through, it presents you with both age options, just like above. Screen Shot 2022-06-23 at 7 52 04 PM

- If you go to the main Albertsons zip code search page and click on the same listing as either of the above ones, it does *not* give you both options! It appears to use your search criteria from the previous page to narrow things down: Screen Shot 2022-06-23 at 4 52 52 PM

So at current, if a user follows one of these links, they might see more options than the metadata about the link suggests they should have. But if they go to the same listing via the search page (instead of our deep link), they will be presented with fewer options, and those options seem to more-or-less match what we are calculating. Maybe. It’s pretty confusing and hard to test exhaustively.

**So! What I did in the PR (using “infant” in the address to determine what `PfizerChild` means) might be right, but it also might be better to treat `PfizerChild` as meaning *both* types of pediatric Pfizer vaccines.** 🤷

/cc @astonm @dhcole @benmatheson @brendanbabb


Mr0grog:
I might go ahead and merge this sometime Friday just so we have *something* for early pediatric Moderna here, but would still love feedback afterward (and will make changes based on it).

astonm:
Man, what a mess. I probably won't have time to give this a good review before merging, but superficially it seems like a reasonable solution.

Will take a look tomorrow and give you some feedback.

Issue #231, opened 2021-06-16
Keep track of manual provider_location overrides


Mr0grog:
We have some situations where the database has manually entered data for provider locations, and we have to be very careful not to accidentally overwrite it. Instead, we should be tracking a list of manually overridden 

One example of this is that NJVSS has some bad 4-digit zip codes, and they are manually fixed in our database. The NJVSS loader has to be very careful to check for bad zip codes so as not to overwrite what’s in that field in the DB: https://github.com/usdigitalresponse/appointment-availability-infra/blob/ce4c936107d2a449d530d41e7d24b5e32f68dffe/loader/src/sources/njvss/index.js#L402-L407

We could track manual overrides in a new DB table as part of #202, but it might be better (*especially* in the short term) to simply keep them in a file in our source code. Storing them in a source code file means they are easy to see, review, and update (and we multiple eyes on them via PRs). We don’t have any GUI that gives us a similarly clear window into overrides we might store in the database. (Props to @janovergoor for pointing this out.)

PR #735, opened 2022-06-24
Add manual corrections to NJVSS

Mr0grog:
This follows the pattern of corrections we added for Albertsons, and solves issues with a few bad addresses. (This was triggered by changes in #726, which put some of these ongoing address errors a little more in my face — which is a good thing.)


Mr0grog:
This fixes #231! I’d totally forgotten about that issue. 🙃

astonm (approved):
👍🏽

Issue #198, opened 2021-06-11
Add loader to get clean location info from State of New Jersey


Mr0grog:
New Jersey used to combine data about locations from a bunch of funky sources, but now has a nice process involving an Airtable Base where staff & volunteers can gather additional info and correct other info with manual overrides. It would be good to augment and improve our location information by having a loader that draws from that, and just posts location information (rather than availability).

We could connect to their Airtable, but it might be better to just read the output from it at `http://njvss-appointments-public.s3.amazonaws.com/data/airtable/data.json.gz`, which is public and requires no authentication.

This is relatively low priority, since most of our customers depend on us for availability info and not as much for location info. In particular, the State of NJ pulls availability from us on the backend, *then* mixes it with this Airtable info. They don’t *need* us to have it mixed in upstream here, and it would mainly benefit other customers who want to show data in NJ.

Mr0grog:
No longer relevant as we move towards shutdown.

PR #736, opened 2022-06-24
Match NJVSS locations with different providers

Mr0grog:
In https://sentry.io/organizations/usdr/issues/3373843306, we're getting a warning because the NJVSS loader can't match an NJVSS location to one we've previously saved. It turns out this is because it has a different provider (`walmart`) which is pretty reasonable. The real problem is that we were only considering locations with NJVSS as a provider for matching, when we should instead be considering all locations in New Jersey.

### DIFF
--- a/loader/src/sources/njvss/index.js
+++ b/loader/src/sources/njvss/index.js
@@ -240,10 +240,12 @@ function filterActualNjvssLocations(locations) {
  * @returns {Array<object>}
  */
 async function findLocationIds(locations) {
+  // FIXME: a lot has changed about how we manage locations and IDs since this
+  // was written. We can probably just use external_ids for this job now!
   let savedLocations;
   try {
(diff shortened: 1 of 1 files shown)


Mr0grog:
I added a FIXME note about some more potential cleanup here. We also need to resolve #198 at some point, which is pretty tightly tied in here.

loader/src/sources/njvss/index.js
```diff
@@ -240,10 +240,12 @@ function filterActualNjvssLocations(locations) {
  * @returns {Array<object>}
  */
 async function findLocationIds(locations) {
+  // FIXME: a lot has changed about how we manage locations and IDs since this
+  // was written. We can probably just use external_ids for this job now!
```
astonm, line 244: 👍🏽

astonm (approved):
lgtm.

PR #738, opened 2022-06-24
Identify early child age range in Albertsons names

Mr0grog:
Between authoring #734 yesterday night and merging today, the actual data has changed quite a bit. "infant" is no longer the only way that early childhood vaccines are indicated; listings now sometimes say pediatric (age 3-5). This modifies our expressions to match it and loosens some other checks that are no longer applicable.

Fixes https://sentry.io/organizations/usdr/issues/3376638147/

Here’s a quick report-out of how this identifies “infant” vs. “pediatric”: https://gist.github.com/Mr0grog/cf10c09b138a2318ce9c3f6b3ac703be

### DIFF
--- a/loader/src/model.js
+++ b/loader/src/model.js
@@ -50,9 +50,19 @@ const PediatricVaccineProducts = new Set([
   VaccineProduct.modernaAge6_11,
 ]);
 
+/**
+ * `VaccineProduct` values that are for very young children.
+ * @readonly
+ */
+const EarlyPediatricVaccineProducts = new Set([
+  VaccineProduct.pfizerAge0_4,
+  VaccineProduct.modernaAge0_5,
+]);
+
 module.exports = {
   Available,
   LocationType,
   VaccineProduct,
   PediatricVaccineProducts,
+  EarlyPediatricVaccineProducts,
 };
--- a/loader/src/sources/albertsons/index.js
+++ b/loader/src/sources/albertsons/index.js
@@ -33,6 +33,7 @@ const {
   LocationType,
   VaccineProduct,
   PediatricVaccineProducts,
+  EarlyPediatricVaccineProducts,
 } = require("../../model");
 const { ParseError } = require("../../exceptions");
 const { corrections } = require("./corrections");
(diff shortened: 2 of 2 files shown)


Mr0grog:
Major risk worth noting here: I’ve removed the bit of code that actually alerted me to this problem. But it just no longer functions correctly anymore, so I’m not sure we can really keep it.

Mr0grog:
Well, I thought I’d be clever and mitigate the above with this warning:

```js
  // So far, we've never seen early childhood products in the same listing as
  // normal childhood or adult products. Warn if we see that, because it
  // *probably* means something has been parsed wrong.
  const hasAdult = products.some((p) => AdultVaccineProducts.has(p));
  const hasPediatric = products.some((p) => PediatricVaccineProducts.has(p));
  const hasInfant = products.some((p) => EarlyPediatricVaccineProducts.has(p));
  if (hasInfant && (hasAdult || hasPediatric)) {
    warn(
      `Listing ${data.id} has infant AND non-infant products. This is probably wrong.`,
      {
        id: data.id,
        address: data.address,
        drugName: data.drugName,
        parsedProducts: products,
      },
      true
    );
  }
```

But then it found two locations that mix pediatric Pfizer and early pediatric Moderna:

```
Albertsons: Listing 1655998149152 has infant AND non-infant products. This is probably wrong. {
  id: '1655998149152',
  address: 'Edlavitch  - 1529 Sixteenth St NW, Washington, DC, 20036',
  drugName: [ 'PfizerChild', 'ModernaChild' ],
  parsedProducts: [ 'pfizer_age_5_11', 'moderna_age_0_5' ]
}
Albertsons: Listing 1617124108762 has infant AND non-infant products. This is probably wrong. {
  id: '1617124108762',
  address: 'Pfizer Child - Albertsons 1262  - 1300 Airport Freeway, Bedford, TX, 76022',
  drugName: [ 'PfizerChild', 'ModernaChild' ],
  parsedProducts: [ 'pfizer_age_5_11', 'moderna_age_0_5' ]
}
```

So much for that, I guess. 😩 (It’s possible this means the Pfizer entry should be interpreted as `pfizer_age_0_4` instead, but really who knows at this point.)

Anyway. Posting the above for after-the-fact thoughts, or for reference if we need it back later.

Mr0grog:
Minor update here: I dug into the front-end code on the search page, and it just… doesn’t differentiate. You can select *either* ages 3–4 or ages 5–11, and both actually just do the same thing. The code is also covered with special case exceptions if a global variable set by the server says it’s a PfizerChild page, which does not happen on the search screen but does happen on other pages, presumably based on stuff in your session. Bottom line: this is super messy inside mHealthCoach (Albertsons’s COVID appointment platform) as well.

PR #746, opened 2022-06-30
Support Albertsons's new *Toddler vaccine codes

Mr0grog:
Albertsons has added new `ModernaToddler` and `PfizerToddler` vaccine codes. Happily, this fixes some of the madness from https://github.com/usdigitalresponse/univaf/pull/738, so we can confidently identify the different varieties of pediatric Pfizer vaccines (on the other hand, they now appear to be using `ModernaToddler` and `ModernaChild` to mean the same thing, but that's better than one code that means two different things).

Fixes https://sentry.io/organizations/usdr/issues/3389381652 and https://sentry.io/organizations/usdr/issues/3389328866.


astonm (approved):
🙌🏽

Issue #683, opened 2022-06-02
Explore running UNIVAF on Render


Mr0grog:
Because management and deployment of infrastructure is much more straightforward and I am leaving USDR, we’d like to set up an experimental deployment on Render, where USDR staff less familiar with our complex AWS & Terraform setup might be more easily able to manage things.

This is a bit of an experiment — if it doesn’t work well, there’s no reason to keep with it. We’ll also still need to maintain a few things on AWS that Render can’t manage:
- Domain registration for getmyvax.org
- S3 buckets for historical data

This *should* mainly consist of setting up a Render team, adding `render.yaml` file to this repo, and tweaking environment variables on Render.

Issue #303, opened 2021-07-13
Scaling to meet load


jaronoff97:
In this issue I'm going to lay out a few ideas I've had that we could implement so that we can better meet potential higher traffic loads. Each of these ideas will have an impact vs difficulty score with it to assist with priority.

## 1: Queue based database updating
Impact: 7
Difficulty: 5

When assessing our per-endpoint metrics in Datadog, the majority of our traffic comes from calls to the `/update` method. In the span of an hour, we have 130k+ update calls, while the closest endpoint only has ~300 calls. In order to better isolate these update calls, and also separate what needs to be able to update the DB vs what needs to read from it, we can have the loaders publish events to a queue, and a small pool of workers that read from the queue to update the DB.

Having these events on a queue would help with failure scenarios where if a worker is failing, another worker can pick up the messages. In the scenario of a bad deployment, the queue would back up such that we wouldn't lose any of those updates when resolved. If we do this work in parallel with the read replica work already in flight, we would also be able to continue serving all requests to the API. The tradeoff here is availability vs consistency, we are optimizing for availability by allowing the queue to back up and the API to continue serving requests rather than waiting for everything to be healthy again.

Implementing this option is also cost-efficient. We would be able to scale down the API task pool because of the lighter load, and have a smaller ECS service for the worker itself. SQS itself costs $0.40 per million requests, at our current rate of around 88-100 million requests to our endpoints a month, our total monthly expected SQS cost would be $40. If we batch the requests we would probably be able to get our cost down to $5-10 a month. Given we'll be able to scale down the API, I expect the costs should about even out.

## 2: Cut down on logging volume
Impact: 2
Difficulty: 1

This is a small issue, but we should look in to it: currently logging is our greatest cost in AWS. It accounts for 40% of our total spend. A quick glance in cloudwatch shows that most of these are just logging "/update 200". Removing that log alone would probably massively decrease our AWS spend and allow us to potentially use a better logging provider.

## 3: Resize our application's resource requests
Impact: 3
Difficulty: 3

[...]

PR #755, opened 2022-07-11
Add Render blueprint

Mr0grog:
This is a first pass at adding configuration for running UNIVAF on Render.com, and partially covers #683.

If this looks good, my plan is to:
1. Deploy this blueprint directly from this branch.
2. Make any changes Render demands based on errors in the file.
3. Replace the database’s contents with a copy of the current production DB from AWS.
4. If all seems good, merge this and switch Render to using the `main` branch.
5. Let that stew for a few days or a week and see if things work well.
6. If everything is all good, point `*.getmyvax.org` at the Render instance and shut off most of the rest of the AWS stuff.


render.yaml
```diff
@@ -0,0 +1,225 @@
...
+  - name: API Server
+    type: web
+    plan: standard
+    env: node
+    repo: "https://github.com/usdigitalresponse/univaf.git"
+    scaling:
+      minInstances: 1
+      maxInstances: 6
+      targetMemoryPercent: 60 # optional if targetCPUPercent is set
+      targetCPUPercent: 60 # optional if targetMemory is set
```
Mr0grog, line 29: `standard` is 1 CPU and 2GB RAM, which is the same as we have the API service currently configured to use in AWS.

This is also a pretty similar scaling configuration to what we currently have: https://github.com/usdigitalresponse/univaf/blob/fd6076ad7e71046a9097105979221c281fc93324/terraform/auto_scaling.tf#L3-L89

The biggest difference is that, on AWS, we set exact thresholds. On Render, we set an optimal usaage level and it decides where the thresholds are: https://render.com/docs/scaling

render.yaml
```diff
@@ -0,0 +1,225 @@
...
+  # NOTE: This first loader also defines properties for re-use by the others.
+  - &loader-config
+    name: "Loader: CVS SMART"
+    schedule: "0/10 * * * *" # Every 10 minutes
+    startCommand: |
+      cd loader
+      bin/univaf-loader cvsSmart --send --compact --states "AL,AK,AZ,AR,CA,CO,CT,DE,DC,FL,GA,HI,ID,IL,IN,IA,KS,KY,LA,ME,MD,MA,MI,MN,MS,MO,MT,NE,NV,NH,NJ,NM,NY,NC,ND,OH,OK,OR,PA,RI,SC,SD,TN,TX,UT,VT,VA,WA,WV,WI,WY,MH,PR,VI"
+    # Common options used for all crons. (Other configs use `<<` to repeat this
+    # cron job's configuration and override the above, non-common values.)
+    plan: starter
```
Mr0grog, line 71: I went with a smaller plan for the loaders, since we are almost certainly way overprovisioned for them on AWS right now (see “resize our application's resource requests” in #303).

render.yaml
```diff
@@ -0,0 +1,225 @@
+# Configure deployment on Render.com infrastructure.
+# For info on options, see: https://render.com/docs/blueprint-spec
+#
+# THIS IS AN EXPERIMENT. Our primary deployment is still on AWS, but we hope to
+# switch to this if it works well and is not too expensive. See:
+# https://github.com/usdigitalresponse/univaf/issues/683
+
+databases:
+  - name: univaf-db-production
+    plan: pro
```
Mr0grog, line 10: This plan is a little smaller than what we are using on RDS right now. Render’s `pro` is:
- 2 CPUs
- 4 GB RAM
- 96 GB of disk
- <= 97 connections

On RDS, we are using a [`db.m5.large` instance](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.DBInstanceClass.html):
- 2 CPUs
- 8 GB RAM
- No imposed connection limit, but we average around 30 and occasionally spike to around 92.
- We have 250 GB of disk but only need ~30 GB; the overage is from back when the DB was growing out of control and we never shrunk after fixing the issues.

We typically have ~5 GB of freeable memory and ~25% CPU utilization (with spikes every 6 hours to ~60%; this is probably autovacuuming and shouldn’t be a real issue). It *seems* to me that this DB plan is probably all we actually need, but if there are issues we can boost it to `pro plus` (4 CPUs, 8 GB RAM, 256 GB disk, 197 connections) → $95/month vs. $185/month.

Last month we paid $158.48 on RDS ($2.72 for backups, $128.16 for the instance, $27.60 for the disk).

Mr0grog: render-db-plans

Mr0grog:
One MAJOR thing this doesn’t account for is metrics. Render makes it easy to ship logs to DataDog, but not so much for metrics.

We currently use StatsD to send metrics by adding DataDog’s StatsD agent as a sidecar container (https://github.com/usdigitalresponse/univaf/blob/fd6076ad7e71046a9097105979221c281fc93324/terraform/modules/task/main.tf#L18-L63), and it doesn’t seem like Render has support for sidecars when deploying via Docker (also: it doesn’t seem like you can deploy a Docker *image* from a registry — Render always wants to build your image for you, which does not seem like the most correct way to be using Docker, but 🤷).

Some options here:
1. Deploy via Docker instead of as a Node.js application and integrate the datadog agent directly into it. This is actually a little complicated, since it means the container is running multiple processes and therefore needs an init system (e.g. [dumb-init](https://github.com/Yelp/dumb-init)) and some additional mess to handle signals correctly. Plus we have to actually figure out the best way to install the DataDog agent since we can’t just use their standard image.
2. Use DataDog’s HTTP API for custom metrics (https://docs.datadoghq.com/api/latest/metrics/#submit-metrics), which requires JS code changes.

Mr0grog:
UPDATE: right after writing the above, I ran across this example of just running the DogStatsD agent as a separate service and connecting over the network. https://github.com/render-examples/datadog-agent

That said, it seems this setup is not quite ideal: https://community.render.com/t/datadog-docker-example-doesnt-work/1622

But better than the other options I listed above.

astonm:
@Mr0grog I like the idea here of setting things up to run on Render.com, and some logistal questions aside, it seems like you've already pulled together a lot of what needs doing here. A couple questions:

- Did you already confirm that Render's YAML parsing accepts the node anchors and merge keys you're using to avoid duplication around the config for each loader?
- Given the length of the configuration file, even with the merge key stuff, I wonder if it might be easier to understand what's going on by writing code to generate the `render.yaml` file rather than editing it by hand.
- Assuming you were down to generate it, could we generate it from the existing Terraform configuration? I imagine if we make the switch to Render, you may have been planning to delete those files, but I would probably argue we should keep them, and if we do, we'll need to keep them in sync.

Mr0grog:
> Did you already confirm that Render's YAML parsing accepts the node anchors

Yep. This is actually an approach I used for a bunch of USDR’s internal tools when I was getting them more organized at the end of last year. (See also https://community.render.com/t/freeform-area-to-set-up-yaml-anchors/2619 and https://feedback.render.com/features/p/support-for-reusable-configuration-in-renderyaml-or-a-place-to-define-variables)

> I wonder if it might be easier to understand what's going on by writing code to generate the `render.yaml` file rather than editing it by hand.

I guess I’m not sure what I would do here other than just pulling out the re-used properties from the first cron job into a separate place. That would be nice, but doesn’t seem obviously worth the tradeoff that now there’s custom tooling to understand and a more complicated process for editing the file (which file do I edit/which can I not touch? Do I have to run a script before committing? Does a CI job do part of this instead? etc.).

Did you have other ideas about what you’d change in this file with a build process?

> could we generate it from the existing Terraform configuration? I imagine if we make the switch to Render, you may have been planning to delete those files, but I would probably argue we should keep them, and if we do, we'll need to keep them in sync.

Hmmm, I’d live to hear your argument for keeping them; you’re right that I figured it makes the most sense to delete (or rather, delete everything except for stuff we’ll still maintain in AWS, e.g. S3 buckets and Route53 DNS stuff).

I’m not super excited about generating it from Terraform; that seems really complicated (not just parsing, but there’s plenty that doesn’t really match up conceptually, like how Render does autoscaling vs. AWS). This feels like a huge amount of work for pretty low payoff — just keeping things in sync while we are running both (which should be a short timeframe). Is there another reason to generate one from the other I’m missing? (Side note: our budget is only another 280 hours over the course of the next year.)

astonm (approved):
I like the idea of keeping the Terraform configs since they feel like a more general way of specifying the infrastructure needs, at least relative to something Render.com-specific (though obviously, they're AWS specific, which is maybe six of one, half dozen of another).

Adding any sort of preprocessing step to generate `render.yaml` will definitely take more effort, and I take the point that simplicity is better here. So let's keep things as is, and (eventually) delete all things Terraform.

lgtm!

Mr0grog:
Making a temporary S3 bucket for the Render to deployment to use while testing: #758.

Mr0grog:
I am currently making a copy of the production DB, which is taking some time. Will deploy this to Render once that’s ready.

Mr0grog:
Updates:
- Realized my problem was that I forgot to exclude the `availability_log` table from the dump. (This also implies that if/when we do a real switchover from AWS, we should do it ASAP after the logs are written to S3.)
- You can’t use dynamic values from other services in env var groups (I knew this, but wasn’t thinking about it), which makes DB configuration super verbose. We probably want to add support for a `DB_URL` that combines all the values. This should just mean using a string for the Knex `connection` option instead of an object: https://github.com/usdigitalresponse/univaf/blob/ef4c2629229b0db09df24ca53508e5d15b1e53e8/server/knexfile.js#L1-L9
- Rebased this to get the latest code.

Taking a lunch break, will deploy to Render after.

PR #758, opened 2022-07-13
Add an S3 bucket for testing the Render deployment

Mr0grog:
This is a pre-requisite for #755. Since the Render deployment is a test, it needs its own S3 bucket to dump historical data into so it doesn't overwrite the actual production data from our stuff running on ECS.

### DIFF
--- a/terraform/s3.tf
+++ b/terraform/s3.tf
@@ -1,16 +1,26 @@
 # s3.tf
 
+locals {
+  data_snapshot_bucket_names = toset([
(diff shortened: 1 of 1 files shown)


Mr0grog:
Well crap, I tried to be clever, but it looks like Terraform is going to delete and re-recreate the bucket. :(

Re-doing this with just a second copy of the same code instead of using `for_each`.

astonm (approved):

PR #760, opened 2022-07-13
Downsize loaders to 0.5 vCPU, 1 GB RAM

Mr0grog:
We've long known that the loaders are probably overprovisioned. This downsizes them by 50%. Partially covers #303.

See https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_size for valid cpu/memory values and combinations.


Mr0grog:
It’s very possible these could work at an even smaller scale. What we’ve set here is almost certainly fine.

PR #764, opened 2022-07-15
542 compress those logfiles

### LINKED ISSUE #542: Write smaller log files
**Not especially high priority, but would have a positive impact on storage and long-term usage of historical data.**

We write out historical logs from the database every night, and sometimes those files are pretty darn big — anywhere from 1.5 GB to 30 GB. I was recently running some numbers for USDR, and downloading these files is just painful. It’s really the most time-consuming part of the process (even from inside the same AWS datacenter where the S3 bucket lives). It would be nice if they were smaller.

Some quick experiments I tried with the log from 2021-09-21, which is 2.6 GB:

| Compression | Size (MB) |
| ----------- | ---: |
| *None*      | 2,600 |
| SQLite | 2,600 |
| GZip | 312 |
| Brotli (quality=5) | 276 |
| GZip each line | 2,200 |
| GZip in chunks of 1,000 lines | 315 |
| GZip in chunks of 10,000 lines | 312 |
| Brotli in chunks of 10,000 lines **(broken)** | 298 |
| Parquet | 461 |
| Parquet using a manual schema | 446 |
| Feather (Apache Arrow) | 735 |
| Feather (Apache Arrow) using a manual schema | 659 |

- I was surprised SQLite didn’t really compress this at all. I think that’s because it doesn’t have special storage type for JSON like Postgres has; under the hood it’s just `TEXT`.
- Brotli at quality 5 is both faster and smaller than GZip at max quality, but…
- I recently read about a neat property of GZip: you can concatenate a bunch of GZip files/binary streams together and the whole thing is a valid, decompressable GZip file! I used this to try GZipping each line and also batches of lines (1,000 lines at a time or 10,000 lines at a time).
    - Line-by-line doesn’t get us much. I think it’s basically only helping for entries with a lot of `capacity` or `slot` entries.
    - On the flip side, doing it in batches is almost as good as doing the whole file!
    - It doesn’t seem like this behavior works for Brotli, though — I tried it and the output couldn’t be read back out and decompressed. 😢
- Parquet and Feather are both columnar formats for dealing with big data, so are pretty well designed for this use case. (I tried these both with `pyarrow`, the Python implementation of Apache Arrow.)
    - Parquet is more storage-optimized, and it shows!
    - Feather is read-optimized (it’s designed for memory-mapping, so smart filtering and aggregation tools can seek to the right parts of the file and do calculations without copying anything, meaning working with the files, once written, is super fast). It definitely results in larger files, but still well-compressed vs. the raw JSON we currently write.
    - Specifying a schema to use improved compression for both of these a little bit. The savings here probably come mostly from making the timestamps numeric and using smaller numbers where appropriate (e.g. `int16` for most counts, since we aren’t ever going to have millions of appointments available at any one location).

Parquet and Feather/Arrow files are really easy to work with in tools like numpy/pandas/dask, which can be a nice advantage. Lots of big data querying tools have native support for them, too. When you don’t have specialized tools, though, they can be a bit of a pain. Both files can work reasonably well for streaming output, although maintaining special schema files for all the tables we output isn’t the most fun thing. Since the core problem here is compression, Parquet is *probably* better than Feather/Arrow.

GZipping in batches works well for streaming output, doesn’t require any specialized tools (Python even has a built-in gzip support that can read line-by-line with low overhead, ideal for this situation, and an ALMOST drop-in replacement for where [appointment-data-insights](https://github.com/usdigitalresponse/appointment-data-insights) uses `ndjson`).

**GZipping in batches of 1,000 lines is probably the best solution here.** (Easily the best compression vs. memory balance here.)

---

For reference, the schema I used:

```py
import pyarrow as pa
log_schema = pa.schema([
    pa.field('source', pa.string()),
    pa.field('valid_at', pa.timestamp('ms', tz='utc')),
    pa.field('checked_at', pa.timestamp('ms', tz='utc')),
    pa.field('changed_at', pa.timestamp('ms', tz='utc')),
    pa.field('available', pa.string()),
    pa.field('meta', pa.string()),
    pa.field('is_public', pa.bool_()),
    pa.field('location_id', pa.string()),
    pa.field('available_count', pa.int16()),
    pa.field('products', pa.list_(pa.string())),
    pa.field('doses', pa.list_(pa.string())),

    pa.field('capacity', pa.list_(pa.struct([
        pa.field('date', pa.timestamp('ms', tz='utc')),
        pa.field('available', pa.string()),
        pa.field('available_count', pa.int16()),
        pa.field('unavailable_count', pa.int16()),
        pa.field('products', pa.list_(pa.string())),
        pa.field('dose', pa.string())
    ]))),

    pa.field('slots', pa.list_(pa.struct([
        pa.field('start', pa.timestamp('ms', tz='utc')),
        pa.field('end', pa.timestamp('ms', tz='utc')),
        pa.field('available', pa.string()),
        pa.field('available_count', pa.int16()),
        pa.field('unavailable_count', pa.int16()),
        pa.field('products', pa.list_(pa.string())),
        pa.field('dose', pa.string()),
        pa.field('booking_url', pa.string())
    ]))),
])
```

Mr0grog:
This gzips our nightly log files before writing them, resulting in drastically smaller and easier to manage data files. At the moment, our data is big enough that the bottleneck is pretty much always bandwidth to load the file (even when loading from disk and not the network!). Gzipping them makes sending, loading, and processing historical data *much* faster. Fixes #542.

The implementation here is based on experimentation and refinement done a while back in https://gist.github.com/Mr0grog/1f5e14ac360e64304c0031282e0cae3f. See that gist and #542 for more details about the approach and other alternatives.

I made this always gzip streams, but we could also change it up so that’s an option.

I also updated a bunch of things here to use the new `compose()` and `pipeline()` builtins here, which forward errors along so we actually handle them correctly in ways that `pipe` does not (luckily we haven’t encountered any weird errors, so this is more just imposing some better practices). However, it’s worth noting that `compose` is still classified as experimental (because it also does fancy stuff like convert generators to streams — very nice). I also didn’t do this in all cases because `compose` can’t take `JSONStream.stringify()` as the last argument (I think it’s just a very old-school first-generation stream, and compose doesn’t support those). Anyway. This is fine, but I think we could also replace the third-party library with:

```js
const ndjsonStream = stream.compose(async function * stringify(source) {
  for await (const object of source) {
    yield JSON.stringify(object);
  }
});
```

🤷

### DIFF
--- a/server/scripts/availability_dump.js
+++ b/server/scripts/availability_dump.js
@@ -23,6 +23,8 @@ const knexConfig = require("../knexfile");
 const luxon = require("luxon");
 const JSONStream = require("JSONStream");
 const stream = require("stream");
+const { Buffer } = require("buffer");
+const zlib = require("zlib");
 
 Sentry.init();
 
@@ -57,6 +59,80 @@ function removeNullPropertiesStream() {
       callback(null, record);
     },
   });
+}
+
+/**
+ * A stream that emits buffers of a given size (in bytes). Useful when an input
+ * stream might emit data in small chunks, while the next stream in a pipeline
+ * works more efficiently with larger chunks, or chunks of a given size.
+ *
+ * In practice, we use this for gzipping, since gzip stream work most
+ * efficiently if data comes in chunks of at least 32 kB (this actually makes
+ * a pretty big difference).
+ *
+ * We couldn't find a good version of this on NPM (which seems surprising,
+ * we're probably missing it). The `stream-chunker` package, which is popular,
+ * performs *terribly* and is often worse than no chunking at all.
+ */
+class BufferedStream extends stream.Transform {
+  constructor({ size = 256 * 1024, ...options } = {}) {
+    super(options);
+    this.size = size;
+    this.resetBuffer();
+  }
+
+  resetBuffer() {
+    this.buffer = Buffer.allocUnsafe(this.size);
+    this.offset = 0;
+  }
+
+  _transform(input, encoding, callback) {
+    if (typeof input === "string") {
+      input = Buffer.from(input, encoding);
+    } else if (!(input instanceof Buffer)) {
+      callback(
+        new TypeError(
+          `BufferedStream input must be strings or buffers, not ${input.constructor.name}`
+        )
+      );
+      return;
+    }
+
+    let inputPosition = 0;
+    while (inputPosition < input.length) {
+      const written = input.copy(this.buffer, this.offset, inputPosition);
+      inputPosition += written;
+      this.offset += written;
+
+      if (this.offset === this.size) {
+        this.push(this.buffer);
+        this.resetBuffer();
+      }
+    }
+    callback();
+  }
+
+  _flush(callback) {
+    if (this.offset > 0) {
+      this.push(this.buffer.slice(0, this.offset));
+    }
+    callback();
+  }
+
+  _destroy(error, callback) {
+    this.buffer = null;
+    callback(error);
+  }
+}
+
+function bufferedGzipStream(size = 64 * 1024) {
+  return stream.compose(
+    new BufferedStream({ size }),
+    zlib.createGzip({
+      level: zlib.constants.Z_BEST_COMPRESSION,
+      chunkSize: size,
+    })
+  );
 }
 
 function getQueryStream(queryBuilder) {
@@ -157,7 +233,7 @@ function formatDate(date) {
 }
 
 function pathFor(type, date) {
-  return `${type}/${type}-${formatDate(date)}.ndjson`;
+  return `${type}/${type}-${formatDate(date)}.ndjson.gz`;
 }
 
 function eachDayOfInterval({ start, end }) {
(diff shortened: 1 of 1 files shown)


server/scripts/availability_dump.js
```diff
@@ -59,6 +61,80 @@ function removeNullPropertiesStream() {
...
+ */
+class BufferedStream extends stream.Transform {
+  constructor({ size = 256 * 1024, ...options } = {}) {
+    super(options);
+    this.size = size;
+    this.resetBuffer();
+  }
+
+  resetBuffer() {
+    this.buffer = Buffer.allocUnsafe(this.size);
```
astonm, line 86: nit: If I correctly understand the (lack of) guarantee `allocUnsafe` makes–that the buffer might already contain data, so you should be careful with reads–we only ever need to allocate a new one here if `this.buffer` is `null` or `undefined`.

astonm (approved):
Cool!

server/scripts/availability_dump.js
Mr0grog: Oh, good point! This really could just happen once in the constructor, and would be more efficient. That leaves the `this.offset = 0;` line as the only thing in the `resetBuffer()` method, so we might as well just drop the method.

I’ll do that before merging later today.

PR #768, opened 2022-07-19
Actually deploy updates to daily-data-snapshot

Mr0grog:
This is a follow-on to #542. Apparently we weren't telling ECS what image tag to use for the daily-data-snapshot task, so we've never actually deployed new code to it. 😱

(I noticed this only because the run today did not write a `.gz` file!)


astonm:
Oof, good catch.

Issue #218, opened 2021-06-16
Add ability to disable loaders


astonm:
During data migrations and other planned API outages, the large number of requests the loaders send to the server can be a nuisance. Currently we can somewhat hackily disable loaders by changing their scheduled run rate to something rare (e.g. `rate(1 day)`) but it would be nice to be able to disable them in a more straightforward way.

Mr0grog:
Note: this functionality is built-in on Render (although you have to do it through the web console instead of `render.yaml`), so #683 results in us switching platforms, this issue will be solved.

PR #778, opened 2022-07-25
Use same DB connection pool for data and logs

Mr0grog:
We used to create two connection pools to the database (one for data and one for writing availability update logs). This leads to a *lot* of connections, and on the new Render deployment we are sometimes running out of connections and dropping logs. This re-uses the same pool for both, which should be safe (we still keep separate references, so the availability log can be changed to point somewhere else) and solves the issue by using half as many connections.

Another option here is to add an environment variable for the maximum number of DB connections in the pool(s). That might still be a good idea, and could be useful in addition to the change I made here, but skipping it seemed simpler for now.

Fixes https://sentry.io/organizations/usdr/issues/3155794613.

@astonm do you mind giving me a gut-check on whether this seems reasonable? The biggest concern I can think of here is that this might reduce throughput slightly, but I *think* it should be low impact overall.

PR #779, opened 2022-07-25
Switch production to Render, tear down deprecated AWS resources

Mr0grog:
⚠️ **DO NOT MERGE! This PR is meant for illustration/planning/discussion purposes. In practice, we’ll want to do this over several steps and not one big PR.** ⚠️

This PR plots a roadmap for completing the switch to Render. We aren’t 100% ready to make the transition yet (e.g. some solution for the issues in #778 needs to land), but I want to queue up something for people to look at and discuss here.

I think the goal is to answer two questions:

1. **Do we want to switch?** The main reason for doing so is making this easier to grok and maintain, and I need feedback from others (@alexallain and, if you have time, @astonm would love your feedback). To answer this, I think it’s useful to look at the changes as a whole (via the “files changed” tab for this PR) or to just look at [the code as it stands this branch](https://github.com/usdigitalresponse/univaf/tree/aws-away) and paying attention to the root-level `README.md`, the `docs/runbook` directory, and the `terraform` directory.
    - Is this an improvement in affairs? Is it more understandable and/or easier to maintain vs. what exists today?
    - We can’t 100% close out AWS, which means this actually leaves us with a *new* service, rather than trading one for the other. Is that OK? Do we need to work harder to clear things out of AWS (e.g. by transferring DNS to USDR’s internal/ops AWS account)?
    - Are there more things we should do or do differently in Render?

2. **Is this a good approach to doing it?** I’ve tried to factor this PR into a series of commits that roughly matches what we’d do and in what sequence. This is *not* a PR that should merge — instead, we’ll want to do each of these steps one by one. Are these the right steps, and in the right order?

    1. Point domains to Render, effectively making it the production deployment. No normal human-friendly URL will point to anything running in AWS (except S3).
        *Note: this is the only step needs to happen at a particular __time__ — immediately after the nightly data snapshot job.*
        *Note 2: This needs to be immediately followed by updating the environment variables for the daily data snapshot cron job so it has permission to write to the right S3 bucket. We’ll have a 23 hour window to do it.*

    3. Update runbook/docs. Should probably happen with the above change.

    4. Disable scheduled tasks (nightly snapshots, pulling from external data sources, e.g. Walgreens). Should happen within a day after (1).

    5. Remove deprecated resources in AWS. This is the *big* change that deletes lots of stuff. It’ll be much more effort to switch back to AWS after this, so we may want to let things bake for a few days or a week beforehand if we are being extra cautious. *Note: this keeps the database, logs, and bastion (bastion is needed to access the database) in case we want to have them as an emergency backup for a while.*

    6. Stop building and pushing docker images. Should probably happen with the above change.

    7. Delete the database in AWS. This is sort of a point of no return (we can recreate it from the data in Render if we want to switch back, but we are effectively starting from scratch at this point).

    8. Remove network configuration and bastion server. Should probably happen with or immediately after the above change, since these things have no point without the database.

    9. Delete old logs from API and loaders. The actual logs for these are stored in CloudWatch, and we’ve rarely used them. We don’t have to delete them if we don’t want, but they are ephemeral — the content will disappear after 1 month regardless, at which point there’s no reason to keep the log groups around.

    10. Drop old Docker container registries. We keep our built Docker images in AWS ECR, and the last step is deleting these. They won’t have really been in use after we do (4) and they can always be recreated from any git commit, but like the database, they’re a big chunk of data that is easier to keep around until we are absolutely sure we don’t need them anymore.

Some other thoughts:

- Is there anything I’ve missed here?

- Is there other stuff (besides the problem #778 is trying to address) that needs doing in Render first?

- @astonm suggested it might be worth keeping a copy of the current Terraform configs around so others can reference it for a full-on AWS deploy (e.g. if someone wants to use GovCloud or similar, Render will not be an option). If we think that’s good, I’m imagining starting this by making a copy of everything currently in `/terraform` and putting it in `/terraform/deprecated` with a little README about its non-maintained status.

- This leaves DNS in the UNIVAF AWS account. Do we want to move it to USDR’s internal/ops AWS account? Upside: it lives where other USDR domains are registered. Downsides: it might be a messy project, it makes management of it invisible to team currently working on UNIVAF (I guess that’s just me 🤷).

- This leaves us with only two really simple things in AWS (DNS and S3). Do we still want to use Terraform to manage them at all? It’s nice that we are managing them in code (and now with no exceptions!), but Terraform is a whole other service we are using to configure AWS (OTOH, it is free, so maybe not a big deal).

---

### Checklist of Actual Transition

- [x] Point CloudFront origin to Render, effectively making it the production deployment. No normal human-friendly URL will point to anything running in AWS (except S3 and CloudFront). (#930, #932, #933, #934, #936, #937)

    *This turned out to be surprisingly complex! Render needs DNS pointing directly at it (not through a reverse-proxy like CloudFront), so this involved a few false starts and a lot of fancy steps with DNS and waiting on various pieces of infrastructure to do their thing. I also double-checked that more Cloudflare-friendly headers wouldn’t cause the built-in Cloudflare DDOS protection on Render to also do some caching (answer: it doesn’t) in #935.*

- [x] Update environment vars/config on Render to be production values. (#930)

- [x] Disable nightly data dump to S3 (in AWS, not Render). (#939)

- [x] Update runbook/docs. (#940)

- [x] *Disable* loaders/all scheduled tasks once Render has been the production service for 1+ days without incident. (Morning 2022-09-15) (#941)

- [x] Stop building and pushing docker images. (#942)

- [x] *Remove* deprecated resources/services in AWS (keeps the database, logs, and bastion (bastion is needed to access the database) in case we want to have them as an emergency backup for a while). This should probably happen a few days after disabling the loaders (above). (Either 2022-09-16 or 2022-09-19)
    - #948
    - #949
    - #950

- [x] Delete the database in AWS. (#951)

- [x] Remove network configuration and bastion server. (#985)

- [ ] Delete old logs (in CloudWatch) from API and loaders.

- [ ] Drop old Docker container registries.

- [ ] Clean up anything that remains in Terraform/AWS.


render.yaml
```diff
@@ -107,6 +107,9 @@ services:
     startCommand: |
       cd server
       node scripts/availability_dump.js --write-to-s3 --clear-log
+    domains:
+      - getmyvax.org
+      - www.getmyvax.org
```
Mr0grog, line 112: I *think* this has to be here otherwise new deployments will blow domains we configure in the web console away. Not 100%, though. @alexallain, we could pilot this by configuring `univaf.usdigitalresponse.org` in the web console and seeing if it survives without explicit configuration in `render.yaml`. (You’d need to create the DNS records in USDR’s Cloudflare account.)

README.md
```diff
@@ -33,7 +33,7 @@ At the top level of this repo, you’ll also find some other useful directories:
 
 - **`scripts`** contains scripts for managing releases, deployments, and other tasks.
 
-- **`terraform`** contains Terraform configuration files used to deploy this project to AWS. We try to keep as much of our infrastructure configuration as possible stored here as code.
+- **`terraform`** contains Terraform configuration files for the parts of this project hosted in AWS (historical data in S3, domain names). We try to keep as much of our infrastructure configuration as possible stored here as code. (Most code runs on [Render](https://render.com) rather than AWS, and is managed in [`render.yaml`](./render.yaml).)
```
astonm, line 36:
```suggestion
- **`terraform`** contains Terraform configuration files for the parts of this project hosted in AWS (historical data in S3, domain names). (Most code runs on [Render](https://render.com) rather than AWS, and is managed in [`render.yaml`](./render.yaml).)
```

astonm (approved):
I haven't given this as deep of a think-through as I'd like yet, but overall, this looks good to me. The sequencing of the PRs to apply makes sense as a rough devops plan as well. 👍🏽 

My gut says there are some unknown unknowns, but I don't see any obvious issues that might arise in the migration. I'll keep noodling on it to see if any come to mind, though. One small thought: Is Render's cron scheduling,  solid? If it's flaky at all, maybe we should look into adding some monitoring for late or missed job runs.

In terms of where we ultimately land on AWS related stuff, I might push for us to go ahead and let it all go. I don't know much about USDR's ability to manage DNS, but unless we are anticipating major changes, I don't see a strong reason to keep the domain configuration in Terraform. S3 makes more sense to maintain alongside the rest of the codebase, but my guess is that we are even less likely to tweak things there in the future than we will with domain names. It doesn't bother me much at all to imagine handling both by just having someone click around on a web interface, and the upside is that we can make the entire `terraform` directory legacy (maybe at `deprecrated/terraform` instead of `terraform/deprecated`).

Mr0grog:
> My gut says there are some unknown unknowns, but I don't see any obvious issues that might arise in the migration.

Yeah, I definitely expect something I’ve done here won’t actually work quite like I expect in practice.

> Is Render's cron scheduling, solid? If it's flaky at all, maybe we should look into adding some monitoring for late or missed job runs.

Yep, it’s pretty solid. I’ve been checking in every so often and things seem to be running pretty reliably and on-schedule. A lot of USDR’s internal ops run as cron jobs on render (often as frequently as every 5 minutes, like we do here), and the launching of those jobs was extremely reliable during the time I was helping with that stuff. That said, we *did* use https://healthchecks.io to monitor some of those jobs, but that was more because a lot of the jobs were failure-prone than because there we issues launching them on-time. We could do the same here if you think it’s a big risk.

> I don't see a strong reason to keep the domain configuration in Terraform. S3 makes more sense to maintain alongside the rest of the codebase, but my guess is that we are even less likely to tweak things there in the future than we will with domain names.

Yeah, I think that seems reasonable. I think I definitely want to hear from Alex and whoever USDR’s new internal ops load is (assuming they’ve started by now) though. But definitely agree if we moved DNS out of the UNIVAF AWS account, keeping Terraform for S3 seems less and less worthwhile the more I think about it.

Mr0grog:
If we do these steps and switch to Render, we should make sure to close #218 after.

Mr0grog:
Rebased to keep this up to date with `main`.

Mr0grog:
Revisiting the first question from this PR, because I don’t think we’ve fully written it out anywhere yet: **Do we want to switch?**

👍 Pros:
- Lower complexity. This is the big one!
    - This branch deletes 1,740 lines of code, most of which are Terraform configuration. Besides that being a lot of code, there’s also plenty of room to get things wrong configuring for AWS since it’s so detailed. Render just gives us fewer opportunities to make mistakes. Terraform configuration also involves learning a whole new language! Render is just one YAML file.
    - In addition to less configuration code, deployment *just happens* on every relevant commit to `main`, rather than requiring other complex bits of process. For example, our Terraform deployment requires a special commit for each deploy (e.g. https://github.com/usdigitalresponse/univaf/commit/2143bf5107c16d5071a654f6b6a45bfbea83affe) and a GitHub actions workflow to create that commit. (Having explicit deploy commits and hashes gives us some nice guarantees and inspectability, but that are really useful in large teams or where more auditability is needed, neither of which really applies to UNIVAF in its current state.
    - Render’s primitives match our application model pretty well. In particular, there are cron jobs, which we have to build for ourselves on AWS by configuring ECS tasks and CloudWatch events.

- It’s much easier to see what’s running, what things are costing, and in general get a handle on the infrastructure in Render.

- Eliminates need for bastion server (sort of). Render lets us spin up ad-hoc shells or SSH sessions on demand, which fulfills most of what we need and use the bastion server for in AWS. Because you *only* get this on-demand, that’s one less piece of infrastructure that can be a security risk. (We could also turn the bastion server in AWS off whenever we don’t use it, but it’s a lot more complicated to do, so we usually don’t bother.)

- Lower costs. The current estimated bill this month from Render is $231.99, while the estimated AWS bill is $736.97. That’s nice, but probably not a huge deal since the cost difference isn’t much next to the cost of having a human on call to maintain the application (whether that’s me or some USDR staff).

    Some other interesting $$$ notes:
    - We spent $60.84 on S3 last month, and that cost will stick around even if we switch to Render. (We also spent 51¢ on Route53 for DNS, which we might keep paying, too.)
    - There’s a lot of room to trim fat on various parts of AWS, but we haven’t prioritized doing so. It requires a lot more careful thinking, and in some cases more complex configuration. Cost management on Render is easier because the options are few and straightforward (see also lower complexity on Render, above 😉). It’s possible we could make AWS cost-competitive, but it might cost more in person time than we’d save on the resources. 🤷
    - Our DataDog costs have gone up a bit:
        - +$15.30/mo on log ingestion. Render lets you send all your logs for all your services or none. Most of these logs are coming from the database, but turning that off means turning off all logs (which we can do — we are making decent use of custom metrics, but not so much of logs). We were not sending UNIVAF logs to DataDog before.
        - +$18.00/mo to run a DataDog agent to collect and send our custom metrics into DataDog. We don't run a separate agent on AWS, so this is a new cost that comes with Render. (It’s possible we might be able to avoid this, but I’m not sure. See [this question on Slack](https://usdigitalresponse.slack.com/archives/C01S4JWSMT7/p1661815283123649). In any case, it’s not much more than the $14.60 we paid AWS last month for the privilege of allowing DataDog to pull metrics from them, so 🤷.
        - -$7.00/mo for our Fargate metrics on AWS ECS. If we switch off AWS, we’ll reduce our DataDog bill by this much.
        - We have also added $16.50 in custom metrics, but we’d keep incurring those regardless of whether we are using AWS or Render, since we’d keep the service instrumented with those metrics. *(It’s possible this might go down if we turn off one of the services, since we’d have half as many hosts sending their versions of the metric. IIUC, each unique metric+tag set = 1 custom metric, and the host is included in the tag set on all our metrics.)*
    - Weird AWS billing details I hadn’t noticed before this:
        - We paid ~$25 in inter-*zone* (not region) bandwidth costs to AWS last month. Render doesn’t appear to charge for this, although it’s unclear how spread across GCP zones things are on Render, since [that info is not visible to Render customers](https://community.render.com/t/any-way-to-determine-the-zone-of-a-render-service/689).
        - We paid $14.60 last month to get metrics data out of AWS and send it to DataDog (at least I’m fairly certain that is the only thing using the CloudWatch GetMetricData API). I hadn’t even considered the fact that we get double-charged on that. 😳

👎 Cons:
- Can’t send all our metrics to DataDog. The Render deployment is still sending all our custom metrics to DataDog, but there doesn’t appear to be a way to send a machine’s CPU/Memory/etc. metrics, [except for the database](https://render.com/docs/datadog#setting-up-postgres-monitoring) (I don’t know why only databases have this level of DataDog support). We can still see these metrics in the Render dashboard, but we just can’t put them alongside our other metrics in a DataDog dashboard.

- Adds another external service to the mix. We can’t move entirely off AWS because we still need our S3 bucket (or something similar) for saving historical data. Render has no built-in equivalent.

- Less than ideal security/permissions. We’ve discussed this elsewhere, but [Render does not have very fine-grained permissions](https://feedback.render.com/features/p/permissions-within-teams). OTOH, I am the only active maintainer right now, which means I’ll have complete admin permissions on AWS, too, so this may be more of a theoretical than real concern.

- Cloudflare. I’m not excited that Render is tied to Cloudflare ([their policies are ethically problematic](https://time.com/6208828/cloudflare-misinformation-internet-research/)).

- In a theoretical world where a county/state/national government wanted to take this on, AWS East/West are already FedRAMP authorized, and AWS GovCloud supports (I think) all the parts of AWS we use. Render is *not* FedRAMP authorized, and due to other maturity issues (e.g. the permissions thing), they probably won’t be able to get authorized any time soon.

render.yaml
Mr0grog: Just ran this experiment with @alexallain, and we need to configure `render.yaml` first. (But it was very quick and worked great!)

Mr0grog:
Result of live discussion with Alex:
- This is a go, and seems like an obvious improvement. 👍
- Keep S3 and DNS maintained in Terraform for now.
- Plan to try and get AWS shut down by the end of next week, but there’s no rush.

Mr0grog:
Another note here: for now, we probably want to keep Cloudfront, and have *that* point to the render instance. (Or maybe to univaf.usdigitalresponse.org?) Render has built-in Cloudflare DDOS protection, but not any configurable kind of CDN/caching.

Mr0grog:
Added a checklist to the top of this PR to track actual progress on the teardown.

Mr0grog:
Current plan:

-d Monday, 9/19: remove API server and load balancer. (#950)
- Thursday, 9/22-ish: remove docker image publishing from CI (#942)
- Thursday, 9/22-ish: delete the database!!
- Monday, 9/26: delete network configurations, IAM policies, bastion server, any remaining infrastructural bits.
- Thursday, 9.29: delete logs, container registries, do any final cleanup. (Gives me two days to do anything weird here before the end of the month, so we are done before the next billing period starts.)

Mr0grog:
Update: **we are reversing course, and switching back to AWS.**

After a little under a month using Render as our production environment, we’ve encountered fairly serious issues with cron scheduling, but more importantly have had unacceptable support response times (> 48 hours in all cases, even just to get a “we’re looking into it” message). We expected some issues and rough edges going in based on USDR’s previous experience with internal ops tooling and the Grants Tools, but did not expect things as severe as we actually encountered. Those might have been manageable if we could count on a support response of some sort on the same day, but we clearly can not. Lots more details can be found on Slack; in particlar this thread: https://usdigitalresponse.slack.com/archives/C01S4JWSMT7/p1664827200958079.

We got pretty far in tearing down old AWS stuff, so there’s a fair amount to put back:
- [ ] Building & publish docker images
- [ ] Rebuild networks/VPCs
- [ ] Create database
- [ ] Create ECS cluster, load balancers, and API service
- [ ] Create daily data snapshot task
- [ ] Create loaders
- [ ] Verify everything is working
- [ ] Switch CloudFront to point at ECS

During the previous transition, we had `render.getmyvax.org` pointing at Render, and I think we’ll repeat that until we are ready to completely tear things down.

We also have some nice lessons learned in terms of tuning our services and resource usage that we should be able to apply here as we rebuild.

Mr0grog:
Database and bastion both deployed, can connect and do things in DB from bastion. Should be ready to start real services on ECS again.

Mr0grog:
The new (old) ECS deployment is running at https://ecs.getmyvax.org .

Mr0grog:
Production is back to AWS. I still need to update the docs here and do a writeup for USDR’s engineering community, but otherwise this is done.

Mr0grog:
Docs updated in #1057; I’m closing this.

PR #781, opened 2022-07-28
Log changes to database connection pool size

Mr0grog:
While addressing https://sentry.io/organizations/usdr/issues/3155794613, we realized we don't really have a handle on exactly how the database connection pools are behaving (see discussion in #778). This change logs every time the pool grows or shrinks, which should give us a better idea how the two pools are behaving. It'll log lines like:

```
2022-07-28T19:37:00.711Z [univaf] info: squarpix.lan Data DB pool size: 1
2022-07-28T19:37:00.714Z [univaf] info: squarpix.lan Data DB pool size: 2
2022-07-28T19:37:00.715Z [univaf] info: squarpix.lan Data DB pool size: 3
2022-07-28T19:37:31.763Z [univaf] info: squarpix.lan Data DB pool size: 2
```

Note: the hostname is included here (`squarpix.lan` in the example above) because we are usually running multiple instances of the server, and we need to know which instance we are logging pool info for to actually understand things.

Another interesting lesson learned here: Knex's pool will reclaim and destroy connection objects from the pool if not used for a while, so that probably has a lot to do with how unpredictable hitting the connection limits seems to be.

### DIFF
--- a/server/src/db-client.ts
+++ b/server/src/db-client.ts
@@ -1,0 +1,20 @@
+import Knex from "knex";
+import type { Knex as KnexType } from "knex";
+import { logger } from "./logger";
+import { getHostInstance, loadDbConfig } from "./config";
+
+export function createDbClient(label: string): KnexType<any, unknown[]> {
+  const client = Knex(loadDbConfig());
+
+  // Add debug-related logging.
+  const pool = client.client.pool;
+  const instanceName = getHostInstance();
+  function logPoolSize() {
+    const poolSize = pool.numUsed() + pool.numFree();
+    logger.info(`${instanceName} ${label} DB pool size: ${poolSize}`);
+  }
+  pool.on("createSuccess", logPoolSize);
+  pool.on("destroySuccess", logPoolSize);
+
+  return client;
+}
--- a/server/src/db-client.ts
+++ b/server/src/db-client.ts
@@ -7,7 +7,7 @@ export function createDbClient(label: string): KnexType<any, unknown[]> {
   const client = Knex(loadDbConfig());
 
   // Add debug-related logging.
-  const pool = client.client.pool;
+  const pool = (client.client as KnexType.Client).pool;
   const instanceName = getHostInstance();
   function logPoolSize() {
     const poolSize = pool.numUsed() + pool.numFree();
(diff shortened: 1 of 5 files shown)


server/src/db-client.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function createDbClient(label: string): KnexType<any, unknown[]> {
+  const client = Knex(loadDbConfig());
+
+  // Add debug-related logging.
+  const pool = client.client.pool;
+  const instanceName = getHostInstance();
+  function logPoolSize() {
+    const poolSize = pool.numUsed() + pool.numFree();
+    logger.info(`${instanceName} ${label} DB pool size: ${poolSize}`);
```
Mr0grog, line 14: Once we have a handle on this, I‘m going to flip it to `logger.debug`. But for now, actually printing debug logs is a lot noisier than is worthwhile.

server/src/db-client.ts
```diff
@@ -0,0 +1,20 @@
+import Knex from "knex";
+import type { Knex as KnexType } from "knex";
+import { logger } from "./logger";
+import { getHostInstance, loadDbConfig } from "./config";
+
+export function createDbClient(label: string): KnexType<any, unknown[]> {
+  const client = Knex(loadDbConfig());
+
+  // Add debug-related logging.
+  const pool = client.client.pool;
```
astonm, line 10: Looks like a light Law of Demeter violation here. Do you think we should guard against `client.client` or `client.client.pool` not existing?

astonm (approved):
Cool, will be good info to have.

server/src/db-client.ts
Mr0grog: I think this is ok — Demeter-wise, this module exists black-box special things we do to Knex, and is well covered by tests and, partially, the TypeScript compiler (`client.client` is publicly typed, but as `any`, even though `Knex.Client` is publicly exported as a type, which is 🤷… I added a new commit with a typecast to use it).

I‘d rather have the tests and/or TSC scream at us in an upgrade PR (cool thing: Tarn’s typings include the event names, so we can’t have a wrong sting in there, either!) so we know to find a new approach to logging than have it silently stop logging if something changes.

Screen Shot 2022-07-29 at 11 20 19 AM

Mr0grog: Also submitted a bug on Knex for the types: https://github.com/knex/knex/issues/5283

PR #783, opened 2022-07-29
2022-07-28 Update dev deps

Mr0grog:
We’ve missed a handful of major dependency updates since Snyk has not fed them to us. This goes through and updates all development dependencies (need to also do this for runtime dependencies).


server/test/lib.ts
```diff
@@ -36,7 +36,7 @@ export function useServerForTests(app: Application): Context {
       context.server.close((error?: Error) => {
         // Jest needs a tick after server shutdown to detect
         // that the resources have been released.
-        setTimeout(() => done(error), 0);
+        setTimeout(() => done(error), 10);
```
Mr0grog, line 39: Jest 28 seems to have tighter/flakier timing here, and adding 10ms seemed to solve it here. :\

PR #786, opened 2022-07-29
Use gauge instead of logging to track DB conns

Mr0grog:
In #781 I added logging for the size of the database connection pool(s). I realized later that DataDog offers a `gauge` metric that lets us graph and solve this use case much better. https://docs.datadoghq.com/metrics/types/?tab=gauge#definition

### DIFF
--- a/server/src/db-client.ts
+++ b/server/src/db-client.ts
@@ -2,6 +2,7 @@ import Knex from "knex";
 import type { Knex as KnexType } from "knex";
 import { logger } from "./logger";
 import { getHostInstance, loadDbConfig } from "./config";
+import { dogstatsd } from "./datadog";
 
 export function createDbClient(label: string): KnexType<any, unknown[]> {
   const client = Knex(loadDbConfig());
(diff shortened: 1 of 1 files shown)


server/src/db-client.ts
```diff
@@ -11,7 +12,10 @@ export function createDbClient(label: string): KnexType<any, unknown[]> {
   const instanceName = getHostInstance();
   function logPoolSize() {
     const poolSize = pool.numUsed() + pool.numFree();
-    logger.info(`${instanceName} ${label} DB pool size: ${poolSize}`);
+    logger.debug(`${instanceName} ${label} DB pool size: ${poolSize}`);
+    dogstatsd.gauge(`db.${label.toLowerCase()}.pool.size`, poolSize, [
+      `instance:${instanceName}`,
+    ]);
```
Mr0grog, line 18: The biggest question here is: is putting the “label” in the metric name and the instance as a tag correct, or should it be the other way around? Or should both be tags?

astonm: 🤷🏽 I'm not super familiar with Datadog, so not sure at all.

Mr0grog: No worries. I’ll go with what I did here, which is my gut instinct (I *think* this configuration makes it easiest to graph them against each other and to filter to a certain host, but am also no expert on DataDog).

PR #787, opened 2022-07-30
Handle Moderna ages 12-17 listings from HyVee

Mr0grog:
HyVee started listing "Moderna (12-17)" as a vaccine type, but it's not actually a different product from what's used for ages 12+ or 18+. The location in question also lists some younger pediatric vaccines, so presumably this doesn't have anything to do with the minimum age the pharmacist is qualified for, and presumably there are just some appointment slots that are age limited this way.

I've classified these as normal `moderna` here, and that is probably a close enough approximation to make sense, rather than creating a new vaccine type for this weirdness (but if we want to there is the [someone-is-always-doing-something-weird](https://github.com/usdigitalresponse/univaf/compare/0605690...someone-is-always-doing-something-weird) branch). HyVee doesn’t have any locations in NJ or AK, so this is probably less of a big deal anyway.

Fixes https://sentry.io/organizations/usdr/issues/3464812423/

### DIFF
--- a/loader/src/sources/hyvee/index.js
+++ b/loader/src/sources/hyvee/index.js
@@ -24,6 +24,9 @@ const SOURCE_NAME = "univaf-hyvee";
 const PROVIDER_NAME = "hyvee";
 const VACCINE_NAMES = {
   Moderna: VaccineProduct.moderna,
+  // Moderna 12-17 is the same product as 12+ or 18+. This is a little messy,
+  // since technically these might be separate appointment slots for teens.
+  "Moderna (12-17)": VaccineProduct.moderna,
   "Pediatric-Moderna (3-5)": VaccineProduct.modernaAge0_5,
   "Pfizer-BioNTech": VaccineProduct.pfizer,
   "Pediatric-Pfizer (5-11)": VaccineProduct.pfizerAge5_11,


Mr0grog:
Just capturing some background from investigating this here…

This is a funny one because I recall there being different dosages for 12-17 and 18+ and 18+ booster for Moderna, but they are all the same formulation/CVX/NDC. At first I thought I got this wrong and added a new product code in the [someone-is-always-doing-something-weird](https://github.com/usdigitalresponse/univaf/compare/0605690...someone-is-always-doing-something-weird) branch, but after more investigating realized my original understanding was correct.

The CDC has made this a little confusing; some pages don’t list any differences (including the [code listing page](https://www.cdc.gov/vaccines/programs/iis/COVID-19-related-codes.html), which is still the most detailed and specific documentation I know of), but then the [vaccine administration page](https://www.cdc.gov/vaccines/covid-19/info-by-product/moderna/administration.html) lists them separately and makes them sound different (“Red vial cap with blue-bordered label – 12 through 17 years of age” vs. “Red cap – 18 years of age and older”), but further investigation reveals that these are both the same; the 18+ vaccine is also a red cap with a blue border (see the [CDC’s full immunization schedule document](https://www.cdc.gov/vaccines/covid-19/downloads/COVID-19-immunization-schedule-ages-6months-older.pdf) or [Moderna’s information for providers](https://eua.modernatx.com/covid19vaccine-eua/providers/dosing-administration)) 🤷).

At the end of the day, I’m not sure if HyVee is doing separate appointment slots for people under vs. over 18 (and therefore splitting up appointments for the same Moderna product) or if they’ve just wound up doing something weird as a result of the way the CDC is messaging this in their administration guides.

In any case, the only location where this is listed also has normal adult Moderna listed, and I’m not sure the distinction is meaningful in the first place. HyVee also doesn’t have any locations in our current partners’ states, so it probably doesn’t make sense to add a new vaccine code or do anything special to represent this in our system.

PR #794, opened 2022-08-05
Implement Sentry tracing for API server

Mr0grog:
We've had Sentry's tracing tools in our dependencies for a while, but never gotten around to setting them up properly. This sets things up for the server (it doesn't address the loader since there's no built-in stuff and there needs to be a lot of work putting spans in all the appropriate places). There are probably a lot of places where we could put in spans for more detail, but this is a start.


astonm (approved):
lgtm

PR #813, opened 2022-08-09
Use new JS languge features in Node 16

Mr0grog:
There are a number of new niceties we should be taking advantage of in the version of Node.js we are using (and more in 18, when we upgrade).


astonm (approved):
👍🏽

PR #816, opened 2022-08-11
Reduce default Sentry sample rate

Mr0grog:
We hit the transaction limit in Sentry less than a week after I turned on performance monitoring (#794). This drops us down to a much more minimal sample rate (we can adjust back up later as appropriate or if/when we turn off either the ECS or Render deployments.

### DIFF
--- a/render.yaml
+++ b/render.yaml
@@ -257,7 +257,7 @@ envVarGroups:
       - key: SENTRY_DSN
         sync: false
       - key: SENTRY_TRACES_SAMPLE_RATE
-        value: "0.2"
+        value: "0.01"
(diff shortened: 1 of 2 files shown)


Mr0grog:
From an e-mail @alexallain received:

Screen Shot 2022-08-11 at 4 27 19 PM

Mr0grog:
I’m still not sure where to see our limits in the actual Sentry web app, so we were and still are flying a little blind. But if it took us 6 days to hit the limit at a sample rate of 0.2, 0.01 should give us more than enough headroom. I don't know that we’ll see any requests except `/api/edge/update` at that sample rate, though, so it may kind of defeat the point. :\

PR #818, opened 2022-08-15
Make DB pool sizes adjustable by connection type

Mr0grog:
We need to manage our database connections on Render more carefully (the limits are *much* lower), and this gives us the ability to configure the connection pool size for regular DB connections vs availability log connections separately. It also configures the availability log connections to have a smaller pool size (5, rather than 10) since there is usually much lower contention for them.

Fixes https://sentry.io/organizations/usdr/issues/3155794613 and is an alternative to #778, based on @astonm’s advice.

### DIFF
--- a/server/src/config.ts
+++ b/server/src/config.ts
@@ -62,8 +62,29 @@ export function getHostInstance(): string {
   }
 }
 
-export function loadDbConfig(): Knex.Config {
+function getPoolSize(connectionName?: string): number {
+  if (!connectionName) return 0;
+
+  const sizeName = `DB_POOL_SIZE_${connectionName.toUpperCase()}`;
+  const rawSize = process.env[sizeName];
+  if (!rawSize) return 0;
+
+  const size = parseInt(rawSize, 10);
+  if (isNaN(size) || size < 1) {
+    throw new TypeError(`${sizeName} must be > 0 (not "${rawSize}")`);
+  }
+  return size;
+}
+
+export function loadDbConfig(connectionName?: string): Knex.Config {
   const knexfile = require("../knexfile");
   const nodeEnv = process.env.NODE_ENV || "development";
-  return knexfile[nodeEnv];
+  const poolSize = getPoolSize(connectionName);
+  return {
+    ...knexfile[nodeEnv],
+    pool: {
+      ...knexfile[nodeEnv].pool,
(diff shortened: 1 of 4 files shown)


server/src/config.ts
```diff
@@ -62,8 +62,29 @@ export function getHostInstance(): string {
...
-export function loadDbConfig(): Knex.Config {
+function getPoolSize(connectionName?: string): number {
+  if (!connectionName) return 0;
+
+  const sizeName = `DB_POOL_SIZE_${connectionName.toUpperCase()}`;
+  const rawSize = process.env[sizeName];
+  if (!rawSize) return 0;
+
+  const size = parseInt(rawSize, 10);
+  if (isNaN(size) || size < 1) {
```
astonm, line 73: ...but here `0` is invalid and will cause an exception to be thrown.

server/src/config.ts
```diff
@@ -62,8 +62,29 @@ export function getHostInstance(): string {
   }
 }
 
-export function loadDbConfig(): Knex.Config {
+function getPoolSize(connectionName?: string): number {
+  if (!connectionName) return 0;
```
astonm, line 66: I don't mind `getPoolSize` working as is, but a nit: I think it's a little inconsistent that here `0` is a valid return value...

astonm (approved):
Looks good to 🚢 , with small thoughts below.

server/src/config.ts
astonm, line 66: I don't mind `getPoolSize` working as is, but a nit: I think it's a little inconsistent that here `0` is a valid return value...

server/src/config.ts
astonm, line 73: ...but here `0` is invalid and will cause an exception to be thrown.

Mr0grog: Hmmm, I guess I was thinking:
1. It’s helpful if this stops you from doing something funky (did you mean to never make any connections? to use the default [then why not specify nothing]? did you make a typo? did we have a parsing problem?).
2. But internally the contract is clearer, and it’s easy enough for `0` to be a non-value for a number-returning function.

But I take your point, it’s probably not that important to protect against 0, and it would be overall more consistent.

Mr0grog: Updated to accept 0.

Issue #377, opened 2021-09-27
Use `http-terminator` for more graceful server shutdown


Mr0grog:
While digging through Express docs for #358, I happened across some discussion of the [http-terminator package](https://www.npmjs.com/package/http-terminator), which actually does quite a bit of useful and complex graceful shutdown behavior (quoted from the docs):

> - it immediately destroys all sockets without an attached HTTP request
> - it allows graceful timeout to sockets with ongoing HTTP requests
> - it properly handles HTTPS connections
> - it informs connections using keep-alive that server is shutting down by setting a connection: close header

This isn’t super high-priority, but we should add http-terminator to our existing graceful shutdown behavior in `server/src/server.ts`: https://github.com/usdigitalresponse/univaf/blob/e8bbdeaf118ac2d817c9d1cecebcaab787999f7a/server/src/server.ts#L13-L21

PR #819, opened 2022-08-15
Graceful shutdown in production

Mr0grog:
We used to shut down the server gracefully on SIGINT, but not SIGTERM. However, SIGTERM is what shuts us down in production, which is arguably the time at which we need to be more graceful! This sets up graceful shutdown for several relevant signals.

It also makes sure we destroy availability logging connections during shutdown (not just normal data connections). This isn't a *huge* deal, since these connections would get reaped pretty quickly when the process dies anyway, but this should make things a little nicer on the DB.

### DIFF
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -22,6 +22,7 @@
         "errorhandler": "1.5.1",
         "express": "4.18.1",
         "hot-shots": "^9.2.0",
+        "http-terminator": "^3.2.0",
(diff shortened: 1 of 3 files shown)


Mr0grog:
Figured I may as well fix #377 while I’m here and use `http-terminator`, which does a bit better job and gives us a timeout.

astonm (approved):
lgtm. And good catch on `availabilityDb` here, too.

PR #834, opened 2022-08-16
Check for and handle errors in HyVee responses

Mr0grog:
It turns out we weren't checking for and surfacing error messages from HyVee's GraphQL responses, which led to this confusing error early this morning: https://sentry.io/organizations/usdr/issues/3505575050. We didn't capture much detail because we didn't even realize the response was an error. Now we surface a better error at a more appropriate point.

This is a standard format for errors in GraphQL, so I've just genericized the handling we already had in the Washington Dept. of Health source.

### DIFF
--- a/loader/src/sources/hyvee/index.js
+++ b/loader/src/sources/hyvee/index.js
@@ -14,9 +14,11 @@
  * not worthwhile at the moment.
  */
 
+const assert = require("node:assert/strict");
 const Sentry = require("@sentry/node");
 const { httpClient, createWarningLogger } = require("../../utils");
 const { LocationType, Available, VaccineProduct } = require("../../model");
+const { GraphQlError } = require("../../exceptions");
 
 const API_URL = "https://www.hy-vee.com/my-pharmacy/api/graphql";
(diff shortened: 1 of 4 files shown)


loader/src/sources/hyvee/index.js
```diff
@@ -99,7 +101,16 @@ async function fetchRawData() {
     },
   });
 
-  return response.body.data.searchPharmaciesNearPoint;
+  if (response.statusCode >= 400 || response.body.errors) {
+    throw new GraphQlError(response);
+  }
+
+  const result = response.body.data.searchPharmaciesNearPoint;
```
astonm, line 108: I know we didn't have it before, but any benefit to guarding against `response.body` or `response.body.data` being undefined? (If so here, same for `response.body.errors` above)

astonm (approved):
:+1:

loader/src/sources/hyvee/index.js
Mr0grog: Not `response.body`, but good point that anything inside it should probably be guarded.

Mr0grog: Should be fixed in f1bdb7afa4e418d4cc3a5dd78e0d308832945d45.

PR #839, opened 2022-08-19
Clean up CLI parsing boilerplate, allow default `--states`

Mr0grog:
This fixes some long-standing bad decisions around parsing CLI options and default values in the loader (all my own fault):
- We had lots of repetitive parsing and error-checking code for converting `--states` to an array and validating other options. These always should have been handled up-front in the CLI parsing code, and they now are.

- Now that the parsing and validation is done in one place, I made it a little more detailed and strict.

- Removed the source-specific state options (e.g. `--rite-aid-states`). I originally made these because I expected we’d call the loader with multiple sources and often need to specify different sets of states for each source. We’ve never used it that way in production, though, so there’s really no value in supporting those options.

- Add/allow default values for `--states`. Not having this has always made ad-hoc usage and testing a pain. Sources now each have their own reasonable default for `--states` that loads all the states that are relevant. (This also means we don’t need to list states in most of the loader configurations in `loaders.tf` and `render.yaml`, but I’ve left them in for maximum clarify/specificity.)

### DIFF
--- a/loader/src/cli.js
+++ b/loader/src/cli.js
@@ -6,6 +6,7 @@ const yargs = require("yargs");
 const { ApiClient } = require("./api-client");
 const { sources } = require("./index");
 const { oneLine } = require("./utils");
+const allStates = require("./states.json");
 
 Sentry.init();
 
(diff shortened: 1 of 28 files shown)


astonm (approved):
Nice, this cleans things up a lot.

PR #840, opened 2022-08-19
Support Hy-Vee's new query format

Mr0grog:
Hy-Vee no longer supports an `isBooster` field in their GraphQL queries. Instead, there is a `doseType` field which has more detailed values. Fixes https://sentry.io/organizations/usdr/issues/3515438646.


Mr0grog:
Worth noting: we don’t actually do anything with this field, but good to know it’s there in case we ever have a reason to.

Mr0grog:
Also: the changes #834 were well timed! They made this much quicker and easier to address. :D

PR #841, opened 2022-08-19
Remove ability to run loader as HTTP server

Mr0grog:
Back when we were working with Vaccinate the States, we implemented the ability to run the loader as a server. You could send it an HTTP POST request with a JSON body equivalent to the normal CLI arguments, and it would load data from the requested sources just as if you had run the loader normally from the CLI. The goal was to make it easy to run on Google Cloud, which has a mechanism for scheduling HTTP requests but not for scheduling scripts or Docker runs.

That service only ran briefly, and Vaccinate the States has since shut down. We no longer use this functionality and maintaining it is probably not a worthwhile burden (it was also never really that well documented, and might be a surprising discovery for anyone new from USDR who might have to do maintenance work here).


loader/Dockerfile
```diff
@@ -21,5 +21,5 @@ ENV RELEASE="${RELEASE}"
 ENV API_URL=""
 ENV API_KEY=""
 
-# run our built server
+# Run the loader
```
Mr0grog, line 24: The previous comment here was never accurate, but I saw it when trying to make sure I’d removed all references to the server and figured I may as well fix it.

astonm (approved):

PR #843, opened 2022-08-22
Clean up PrepMod addresses during formatting

### LINKED ISSUE #458: Some PrepMod locations have full addresses in `address.line`

Mr0grog:
Sometimes PrepMod data has full addresses in the `address.line` field, despite also having the city/state/etc. fields filled in. This attempts to clean that up automatically where possible, and also address issues where multiple lines are conjoined into one in the data, e.g. `Loussac Library / 3600 Denali St`.

Fixes #458.

This doesn’t seem to be happening with any current data in PrepMod (yay), so I tested against some old examples (and then manually cleaned them up in our system). Some manual fixes:

```
   id: a8096272-a874-43a2-8390-7a76f5434a34
 from: [ '30 Mall Drive W. Jersey City, NJ 07302' ]
fixed: [ '30 Mall Drive W.' ]

   id: be266796-f2ce-422b-a138-4f3912e65f25
 from: [ '1 Cedar Crest Village Pompton Plains, NJ 07444' ]
fixed: [ '1 Cedar Crest Village' ]

   id: 0591c0f2-7ffc-4b33-8d6c-172490b4cf12
 from: [ '3000 Essex Road Tinton Falls, NJ 07753' ]
fixed: [ '3000 Essex Road' ]

   id: 24eaaaf9-67ea-4f52-8037-051a945628d9
 from: [ '1306 Esterbrook Ave, Rahway, NJ 07065' ]
fixed: [ '1306 Esterbrook Ave' ]

   id: 0e019272-9229-4cdd-a220-196f8bd17652
 from: [ 'Evergreen Middle School: 7621 Beverly Ln, Everett, WA 98203' ]
fixed: [ 'Evergreen Middle School', '7621 Beverly Ln' ]

   id: 35c33bd5-9c26-4386-bd01-227ce653deba
 from: [ 'Loussac Library 3600 Denali St, Anchorage, AK 99503' ]
fixed: [ 'Loussac Library 3600 Denali St' ]

   id: 59389772-2e4a-4183-93b1-2339ab7eeff0
 from: [ 'Old Lynden Middle School Cafeteria, 516 Main Street, Lynden, WA 98264' ]
fixed: [ 'Old Lynden Middle School Cafeteria', '516 Main Street' ]

   id: 5580fc8d-21a7-4606-93cf-04e5fc6031d8
 from: [ '427 Darrow Ave. Plainfield, NJ 07060' ]
fixed: [ '427 Darrow Ave.' ]

   id: 890783e2-d768-4f29-aad5-90669e73c07e
 from: [ '49th State Brewing Company 717 W 3rd Ave, Anchorage, AK 99501' ]
fixed: [ '49th State Brewing Company', '717 W 3rd Ave' ]

   id: 3c96af5d-ccbf-4072-b785-e14a0ba93b3c
 from: [ '115 E Delaware Avenue, Pennington, NJ 08534' ]
fixed: [ '115 E Delaware Avenue' ]
```

### DIFF
--- a/loader/src/smart-scheduling-links.js
+++ b/loader/src/smart-scheduling-links.js
@@ -59,6 +59,15 @@ const scheduleReference = Symbol("schedule");
 const sourceReference = Symbol("source");
 
 /**
+ * @typedef {Object} SmartSchdulingLinksAddress
+ * @property {string[]} line
+ * @property {string} city
+ * @property {string} state
+ * @property {string} postalCode
+ * @property {string} [district]
+ */
+
+/**
  * Compare two lists of states to see if they include any states in common.
  * If either of the values is `null` or `undefined` they are assumed to match.
  * The second value *may* be a string identifying a single state instead of an
--- a/loader/src/sources/prepmod/index.js
+++ b/loader/src/sources/prepmod/index.js
@@ -174,7 +174,10 @@ function formatAddress(rawAddress) {
   // is still a little speculative, so this will help us evaluate whether it's
   // behaving correctly. Once we are more certain, this can be removed.
   if (!isDeepStrictEqual(cleanLines, extraCleanLines)) {
-    warn(`Poorly formatted address lines: ${JSON.stringify(rawAddress.line)}`);
+    warn(`Poorly formatted address lines: ${JSON.stringify(rawAddress.line)}`, {
+      original: JSON.stringify(rawAddress.line),
+      formatted: JSON.stringify(extraCleanLines),
+    });
   }
 
   return {
--- a/loader/src/sources/prepmod/index.js
+++ b/loader/src/sources/prepmod/index.js
@@ -174,10 +174,14 @@ function formatAddress(rawAddress) {
   // is still a little speculative, so this will help us evaluate whether it's
   // behaving correctly. Once we are more certain, this can be removed.
   if (!isDeepStrictEqual(cleanLines, extraCleanLines)) {
-    warn(`Poorly formatted address lines: ${JSON.stringify(rawAddress.line)}`, {
-      original: JSON.stringify(rawAddress.line),
-      formatted: JSON.stringify(extraCleanLines),
-    });
+    warn(
+      `Poorly formatted address lines: ${JSON.stringify(rawAddress.line)}`,
+      {
+        original: JSON.stringify(rawAddress.line),
+        formatted: JSON.stringify(extraCleanLines),
+      },
+      true
+    );
   }
 
   return {
--- a/loader/src/sources/prepmod/index.js
+++ b/loader/src/sources/prepmod/index.js
@@ -141,7 +141,25 @@ function formatLocation(host, validTime, locationInfo) {
 // Attempts to match "<city>, <state abbreviation> <zip>" so it can be removed
 // from the address lines.
 const nonAddressLinePattern =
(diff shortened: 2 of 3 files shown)


astonm (approved):
👍🏽

Mr0grog:
Current problem I’m struggling with, and which I didn’t fully think through before:

There are now a couple locations that this reformats and warns for:

```
prepmod: Poorly formatted address lines: ["13353 Bel-Red Rd, Ste 105"] {
  original: '["13353 Bel-Red Rd, Ste 105"]',
  formatted: '["13353 Bel-Red Rd","Ste 105"]'
}
prepmod: Poorly formatted address lines: ["7631 212th St SW, Ste 100"] {
  original: '["7631 212th St SW, Ste 100"]',
  formatted: '["7631 212th St SW","Ste 100"]'
}
```

This is going to be noisy. I’m wondering about:
1. Just let it send lots of events but ignore the ones that are OK (like these) in Sentry. Works, but not great because they’ll build up over time. (I also need to make sure these fingerprint differently and don’t all merge into one issue in Sentry, because then we can’t usefully ignore.)
2. These are both finding a line that is the suite/unit number. We could identify that pattern and always consider it OK (that is, move it from the `extraCleanLines` part to the `cleanLines` part). That’s nice, but is also just deferring this same problem until it’s something else that is fine, like `401 Coates Avenue Northwest, Yelm Middle School`, which is an actual (former) location we got from PrepMod.
3. Add a `corrections` module [like what we have for Albertsons](https://github.com/usdigitalresponse/univaf/blob/main/loader/src/sources/albertsons/corrections.js). I’m not sure this should be exactly the same, since PrepMod *appears* to have had some instances of addresses changing (BIG typos) in the past, so we maybe want to correct based on the address rather than based on the ID. So still some questions on the structure of this.
4. Add a list of addresses that are OK/correctable and which we don’t warn for. This would probably be in a separate module kind of like `corrections`.
5. A mix of 3 and 4 — 4 to solve this problem, but also 3 for things like complex instructions in the address (e.g. in the past we had a PrepMod location with the address `Inside Olympia Capital Mall, next to Buckle, closest parking in NW lot, near Shoe Dept. Encore, 625 Black Lake Boulevard Southwest`, and part of that address should be moved ot the `description` field).
6. Something else?

astonm:
Oof. 

Given how ambitious the address parsing already is here, I would be tempted to go for option 2. You're so close to done here, I don't know if it's worth going for options 3 or 4 where you'll need to add (and then maintain) one of these accessory lists. I think option 1 isn't very ideal, and all of the other options would be preferable in my eyes.

Mr0grog:
Good point. I’ll go ahead and do that for now, and solve the more complex 3/4 approach when we need it.

PR #859, opened 2022-08-24
Update Terraform to v1.2.8 (latest)

Mr0grog:
The version of Terraform we’re using is way out of date and pre-1.0. We should fix that.


Mr0grog:
We weren’t using any features that run afoul of the 0.14 → 1.0 upgrade, and there don’t appear to be any obvious new features in the changelogs that we should be taking advantage of (I thought the `one()` function might be, but no), so there’s not much to actually change here other than actually specifying the version we use. I’ll update Terraform Cloud before merging.

PR #860, opened 2022-08-24
Specify correct AWS provider for Terraform

Mr0grog:
We use features from version 4.11 of the AWS provider; if you actually installed 4.6 as specified before, our configuration files would not have worked.

### DIFF
--- a/terraform/api.tf
+++ b/terraform/api.tf
@@ -1,13 +1,13 @@
 # Load Balancer
 
-resource "aws_lb" "main" {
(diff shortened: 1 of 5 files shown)


Mr0grog:
NEVERMIND THAT BUSINESS with the alb → lb renaming. It makes Terraform want to delete and recreate a whole mess of stuff, and that doesn’t seem worthwhile right now (especially with a potential switch to Render in flight).

PR #861, opened 2022-08-27
Allow tests to have transactions

### LINKED ISSUE #463: Support transactions in server tests
In #298, we skipped some tests because our database testing setup wraps each test in a transaction, but the test code needs transactions to close in order to function. (See more in this comment: https://github.com/usdigitalresponse/univaf/pull/298#discussion_r664765210, or #100 where we added utilities to wrap tests in transactions.)

What we should probably be doing is truncating all the application tables after each test (note this is specifically application tables, since we don’t want to blow away PostGIS support tables, etc.). We should probably have a list of tables not to truncate, and otherwise just iterate through all the tables and truncate them after each test that uses `installTestDatabaseHooks`: https://github.com/usdigitalresponse/univaf/blob/fe25a3a94975064d8c1457d94fd9173c85757ae9/server/test/lib.ts#L70-L75

Mr0grog:
Instead of wrapping tests in transactions (and making transactions *inside* the test do nothing), truncate the relevant tables and sequences before each test (not after, so if something goes wrong, that leaves stuff in the DB so you can inspect it).

This also fixes a long-standing issue where Jest would try and run tests in the `dist` directory, which would sometimes lead to incorrect failures (if you hadn’t run a build since changing a test file, the test in `dist` would have the old code and fail).

Fixes #463.

### DIFF
--- a/server/jest.config.js
+++ b/server/jest.config.js
@@ -8,7 +8,7 @@ module.exports = {
   transform: {
     "^.+\\.(ts|tsx)$": "ts-jest",
   },
-  testMatch: ["**/test/**/*.test.(ts|js)"],
+  testMatch: ["**/test/**/*.test.(ts|js)", "!**/dist/test/**/*"],
   testEnvironment: "node",
   reporters: ["default", "github-actions"],
   coverageReporters: ["text", "html"],
--- a/server/test/db.test.ts
+++ b/server/test/db.test.ts
@@ -678,7 +678,7 @@ describe("db.addExternalIds", () => {
   };
 
   describe("creating locations", () => {
-    it.skip("does not allow a colon in the externalId", async () => {
+    it("does not allow a colon in the externalId", async () => {
       const TestLocationInvalidExternalId = {
         ...TestLocation,
         ...updatedData,
@@ -690,7 +690,7 @@ describe("db.addExternalIds", () => {
       const newLocation = await getLocationById(
         TestLocationInvalidExternalId.id
       );
-      expect(newLocation).toBe(undefined);
+      expect(newLocation).toBeUndefined();
     });
   });
 
--- a/server/test/lib.ts
+++ b/server/test/lib.ts
@@ -10,6 +10,15 @@ import { TestLocation } from "./fixtures";
 
 import type { Knex } from "knex";
 
+// Don't clean out these tables before/after tests.
+const DO_NOT_RESET_TABLES = new Set([
+  // From PostGIS.
+  "spatial_ref_sys",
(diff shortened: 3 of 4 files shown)


Mr0grog:
I looked at a couple third-party packages for this ([knex-db-manager](https://www.npmjs.com/package/knex-db-manager) and [knex-tablecleaner](https://www.npmjs.com/package/knex-tablecleaner)), but they were both pretty old (had peer dependencies on Knex < 1.0, and Knex is now at 2.x) had problems with the `CASCADE` issue described in a comment in this code, and one dropped tables instead of truncating them, which is a lot more complicated because it requires us to have a schema to recreate the table. 🤷

astonm (changes requested):
Cool, this is in accordance with what we wanted from the start, right? Glad you found a way to get the right behavior for clearing out and resetting tables.

One ask: can we find a way to make sure we're only ever running this truncation code on the test database?

Mr0grog:
> can we find a way to make sure we're only ever running this truncation code on the test database?

Do you have any deeper thoughts about this? It only runs as part of the tests, and only when a test opts into it (same as with the transactions before). We could gate this on the value of `NODE_ENV` (so it only runs with the configuration that gets you the test DB), but then we'd be in a situation where all the tests could run without the DB cleanup, and if they wind up running against a non-test DB, they are still going to make quite a mess.

astonm:
I think checking `NODE_ENV` would satisfy my worries here. If we asserted the right environment when `installTestDatabaseHooks()` is called, wouldn't that stop tests from running altogether (thus nothing to clean up)?

Mr0grog:
Ah, that makes more sense. For some reason I was stuck on the idea that you specifically meant `resetDatabase()` not all the database tooling for tests together.

Mr0grog:
@astonm Ha, I totally forgot we already do that as part of global setup, where we clear the whole database and recreate it from scratch! https://github.com/usdigitalresponse/univaf/blob/62e113fadc55686cc9985747ad498b9fc5386f9b/server/test/globalSetup.ts#L6-L23

So I think there’s nothing that actually needs to change?

astonm (approved):
Hah, I forgot about that myself... Belt and suspenders would be to add a call to `assertIsTestDatabase()` inside of `installTestDatabaseHooks` on some off chance that code gets called without running through the Jest `globalSetup`, but yeah, I think no need. Good to go here!

Mr0grog:
👍 I think I'll merge this tomorrow and see about following it up with a quick PR to refactor the modules with test support code so they can share code and move them into a subfolder (mirroring the organization of `loader/test/support`).

PR #862, opened 2022-08-29
Reorganize test support code

Mr0grog:
This is a follow-on to #861, and got a little more complicated than I expected.

The main headline here is that the server’s test support/library code is all in its own directory at `server/test/support` instead of mixed in with the tests in `server/test`. I also split up the server test hooks, database test hooks and general utilities from each other.

Finally, the part that got messy was in trying to share code between `globalSetup` and `database-testing` — the testing hooks in `database-testing` need to import the `db` and `availability-log` module, which create database connections at import time (not really the greatest pattern, but its what we’ve got), which means `database-testing` can’t really be imported by `global-setup` without setting up and tearing down a bunch of pointless connections. So I made `database-core` which has all the abstract stuff (clearing data, clearing schemas, asserting for the test DB). It’s a little complicated, but the alternative was dynamically importing the built-in DB connections in the Jest hooks in the `database-testing` module, which is its own kind of mess. 🤷

### DIFF
--- a/server/jest.config.js
+++ b/server/jest.config.js
@@ -12,5 +12,5 @@ module.exports = {
   testEnvironment: "node",
   reporters: ["default", "github-actions"],
   coverageReporters: ["text", "html"],
-  globalSetup: "<rootDir>/test/globalSetup.ts",
+  globalSetup: "<rootDir>/test/support/globalSetup.ts",
 };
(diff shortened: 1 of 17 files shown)


astonm (approved):
Cool, thanks for this. Appreciate that things got a little hairy at one point, but the end result looks nice and easy to understand to me.

Mr0grog:
Sweet! That’s what I was hoping.

Issue #669, opened 2022-05-24
Loader tasks occasionally hang


Mr0grog:
Today I discovered a handful of loaders that had somehow gotten stuck hanging for more than a month. Unfortunately, there’s no ECS-native way to add timeouts to tasks (it’s been a planned feature for 2.5 years, though! https://github.com/aws/containers-roadmap/issues/572), so we should consider doing something to handle this ourselves.

Some options:
- Use the `timeout` command as part of the loader entrypoint. For example, instead of: https://github.com/usdigitalresponse/univaf/blob/a5622335ddf5bff353250133ce9b4eba9806909d/loader/Dockerfile#L24-L25

    ```dockerfile
    ENTRYPOINT ["timeout", "1800", "node", "./bin/univaf-loader", "--send", "--compact"]
    ```

    (May also require installing `timeout` with `sudo apt-get install timeout`)

- Add some JS to the CLI entrypoint that starts a timeout that kills the process, for example:

    ```js
    setTimeout(() => process.exit(1), 30 * 60 * 1000).unref()
    ```

- Something with AWS Step Functions (I didn’t even know this existed) as described here: https://github.com/aws/containers-roadmap/issues/572#issuecomment-805996037

- A CloudWatch alarm that notifies us of jobs that have run too long so we can manually kill them.

The first option seems like the most straightforward to me.

Mr0grog:
We have not seen this issue recur since I added timeouts in #880, so I’m marking it as done.

PR #880, opened 2022-08-31
Ensure all HTTP requests eventually time out

Mr0grog:
Got doesn't have a built-in timeout, so it's possible for an HTTP request the loader makes to hang forever. This sets a default timeout of 2.5 minutes on every request. That's long, but at least makes sure things eventually fail. I think this might help with #669 (the loader occasionally hangs and never stops), although it's possible the root cause is something else.


astonm (approved):
👍🏽

PR #882, opened 2022-09-01
Share common code between loader and server

Mr0grog:
⚠️ **Work in progress; not ready for merging!** ⚠️ 

This is an attempt at making it possible for the loader and server to share code rather than having some stuff copy/pasted. The are 2 core goals:
1. Get rid of duplicated code that can get out of sync/produce subtle bugs.
2. Make getting started and working on the project a little easier.

This adds a third source code/project directory named `common` with a package named `univaf-common` in it, and also switches to using NPM workspaces to manage our dependencies instead of setting up each package independently (running `npm install` in the project root installs common, loader, server, and ui all at once; this also allows us to import common code from `univaf-common` instead of relative file paths). The file structure looks like:

```
/
  - package.json  (lists the workspaces - common, loader, server, ui)
  - common/    (New)
    - package.json
    - tsconfig.json
    - src/
      - index.js
      - jest.js
      - states.json
      - <whatever>.js
  - loader/    (No consequential changes)
    - package.json
    - src/
    - test/
  - server/    (No consequential changes)
    - package.json
    - tsconfig.json
    - src/
    - test/
  - ui/        (No consequential changes)
    - package.json
    - src/
```

The source in `common` is JavaScript (so the loader can use it directly), but has a `tsconfig.json` file so the server can understand it and do autocompletion, automatic typing and building (when you build the server, it will build common for you if it’s out of date), etc. That said, we have some stuff (like the [vaccine constants](https://github.com/usdigitalresponse/univaf/blob/d50ccb052292c968b972d903305d111396492353/server/src/vaccines.ts#L1-L51)) that really leverage TypeScript, and it would be a shame to lose that. So we may have to switch to making this package compiled, and make working on the loader a little more complicated.

I don’t want to move all the shared code possible in this PR. Instead, this should get a working framework in so that moving code to a shared location can be an incremental process. That said, some things that should still happen in this PR or a PR that merges before this one:
- Figure out how to handle sharing code that really leverages TypeScript (like the [vaccine constants](https://github.com/usdigitalresponse/univaf/blob/d50ccb052292c968b972d903305d111396492353/server/src/vaccines.ts#L1-L51))
    - Making the `common` package TypeScript makes working on the loader awkward because you have to know to build `common` first.
    - We could make the loader a TypeScript project (even if all the files stay as JS). As a bonus, we get type checking against our JSDoc comments (which will probably produce some errors from bad or unmaintained comments at first).
    - We could just have a build script in the loader that runs the build script in `common`.
- Update docs.
- Update our Dockerfiles and scripts/workflows that use them. They currently only include their subdirectly, but will now need to include the whole project. *OTOH: maybe we just drop the Docker stuff in #779, and then this isn’t a problem.*
- Set up tests for the `univaf-common` package.
- Prettier and ESLint are still defined per-package, but could be moved to the root level.


common/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "typesVersions": {
+    "*": {
+      ".": [
+        "./dist/src/index.d.js"
+      ],
+      "./*": [
+        "./dist/src/*"
+      ]
+    }
+  },
```
Mr0grog, line 20: Note to self for later: I think there’s a way to put the types inside the `exports` object, but I was having trouble with path resolution when I first tried it. Now that I have paths fixed, it might be worth revisiting. See https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/

common/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "peerDependencies": {
+    "jest": ">=28.0.0"
+  },
+  "devDependencies": {
+    "typescript": "4.8.2"
+  },
+  "scripts": {
+    "clean": "tsc --build --clean",
```
Mr0grog, line 28: `--clean` is a new thing when you use `--build`. Should see about using it in the `clean` script for the server package.

Mr0grog:
Superseded by #1133.

PR #886, opened 2022-09-02
Add new codes for bivalent vaccines

Mr0grog:
This adds new UNIVAF vaccine product names and NDC/CVX codes from CDC to support the new bivalent vaccines from Moderna and Pfizer. For now, it adds 3 products:
- `moderna_v2` (designed for ages 6+, but only authorized for 18+)
- `pfizer_v2`
- `pfizer_age_5_11_v2` (not yet authorized, I think, but there are NDC and CVX codes for it)


Mr0grog:
The main question here is whether these names make sense.
- Is the `_v2` in the right place, or should `pfizer_age_5_11_v2` be `pfizer_age_v2_5_11` instead?
- Is `_v2` the right suffix/modifier here? I also considered `_ba4_ba5` or `_omicron` or `+omicron`, but worried these might get problematic with farther future trivalent vaccines.

Mr0grog:
Quick update: we are now seeing new bivalent Moderna boosters from HyVee. That’ll be in #886, which builds on top of this PR.

Mr0grog:
Alaska folks have voiced a preference for `<existing name>_ba4_ba5`,

astonm (approved):
I don't mind using `v2` as a way to denote this next batch of vaccines. It's sufficiently specific to be understandable, but general enough of a naming pattern that we'll know what to do with later ones.

I do have strong feelings about where to put the `_v2` though. I think we should keep it right up against the brand name in all cases, e.g. `pfizer_v2_age_5_11` and `pfizerV2Age5_11`. I think the core concept is that this is the second rev of the vaccine, not that it's the second rev of a particular age group's vaccine, which is how I read it with the `_v2` at the end. Also, I like having all of the v2's sorting lexographically together, and anyway to my eye it looks better visually. 

All that said, if you prefer it as is, I'm cool if you :ship: it.

astonm:
Oops, I didn't see the `_ba4_ba5` comment while I was writing mine. I don't like it as much as `v2`, but I'll admit it does look wonky in the middle of things, so if we go that way, using it as a strict suffix seems like the right thing.

Mr0grog:
> I don't like it as much as v2, but I'll admit it does look wonky in the middle of things, so if we go that way, using it as a strict suffix seems like the right thing.

Ha! I was just leaning towards moving to to the middle, and was going to switch to `_ba4_ba5` based on Alaska’s feedback. Will cogitate a bit on this, I guess (since there’s no rush because we haven’t seen these in AK or NJ yet).

Mr0grog:
OK, this has started showing up in CDC data now, so I’m just going ahead and doing the more detailed, ugly-in-the-middle apprach:
- `moderna_ba4_ba5`
- `pfizer_ba4_ba5`
- `pfizer_ba4_ba5_age_5_11`

PR #906, opened 2022-09-06
Upgrade server to AWS SDK v3

Mr0grog:
The loader has used v3 of the AWS SDK since its inception; this upgrades the server to match. Some nice upsides include only updating when there are changes to S3, a promise-focused API (instead of having to do extra stuff to get promises), and built-in TypeScript support.

One downside here is that multipart uploads got more complicated. There’s still first-party tooling, but it’s in a separate package from S3. It also appears that v3 no longer reads the `AWS_DEFAULT_REGION` environment variable, which we’ve been relying on (at least it didn’t work for me when testing). It looks like we should be setting `AWS_REGION` now instead. For the short term, I shimmed in support for the old variable.

### DIFF
--- a/loader/package-lock.json
+++ b/loader/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.1.0",
       "license": "MIT",
       "dependencies": {
-        "@aws-sdk/client-s3": "^3.163.0",
+        "@aws-sdk/client-s3": "^3.165.0",
         "@sentry/node": "^7.12.1",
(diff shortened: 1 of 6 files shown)


astonm (approved):
👍🏽

Mr0grog:
Was low-key hopping the `queueSize` of 10 might speed this job up, but it doesn’t seem to have done so in tonight’s run. C’est la vie. 🤷

PR #907, opened 2022-09-06
Support DB_URL for tests

Mr0grog:
This is a follow-on to #902. I tested that change by running the server, but never ran tests with `DB_URL` defined! It, of course, does not work without additional support. D'oh!


astonm (approved):
❗

Mr0grog:
See also: why having too many ways to do a thing is not always good. :\

PR #908, opened 2022-09-07
Try downsizing Render database, take 2

Mr0grog:
As expected, #905 did not work. I’ll have to manually create a new database to test this.

### DIFF
--- a/render.yaml
+++ b/render.yaml
@@ -8,7 +8,7 @@
 databases:
   - name: univaf-db-production
     region: oregon
-    plan: standard plus
+    plan: pro
     databaseName: univaf_db_production
     user: univaf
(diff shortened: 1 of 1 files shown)


Mr0grog:
OK, so what I did was disconnected the blueprint on the `main` branch, manually created a smaller DB, imported a fresh dump of production data, updated the `render.yaml` on this branch, and added this branch as the blueprint. Everything sync’d up to the new DB just right.

I’ll disconnect the blueprint again, merge this, and set `main` back as the blueprint, and everything should be good to go.

(Bonus: I set the new DB to send metrics to DataDog. You can only do that when creating a new DB, because Render is not yet a super mature platform.)

PR #911, opened 2022-09-07
Ensure Sentry releases are set on Render Loader

Mr0grog:
Use `RENDER_GIT_COMMIT` to set the release tag for Sentry, just like we do in the server. Ideally I’ll improve this in #882, but for now, this just needs to be added to the loader.


Mr0grog:
I also added in CLI arguments as context, which I’ve been meaning to do forever, since I often find it would be helpful in quickly identifying things:
Screen Shot 2022-09-07 at 4 52 49 PM

PR #915, opened 2022-09-12
Add more metrics to API & DB operations

Mr0grog:
This adds a few metrics (and tags for metrics) to track:
- `api.received.updates.count` are the number of calls to the `update` endpoint, tagged by source.
- `api.received.updates.bytes` (histogram) is the uncompressed body size of each update, tagged by source.
- `db.locations.created.count` is the number of locations created in the DB
- `db.locations.updated.count` is the number of locations updated in the DB
- `db.availability.created.count` is the number of availability records created in the DB
- `db.availability.updated.count` is the number of availability records updated in the DB

All the above have a `source:<source_name>` tag, and request metrics now also have a `internal:<true|false>` tag for whether the request came from part of UNIVAF, or from someone else (e.g. state of NJ).

I prototyped some code to track request and response size to every request, but wasn’t sure it was worthwhile, and left it out for now. (We currently get some basic metrics on this from our load balancer in AWS, but don’t have access to that data from Render.)

### DIFF
--- a/server/src/db.ts
+++ b/server/src/db.ts
@@ -18,6 +18,7 @@ import * as Sentry from "@sentry/node";
 import * as availabilityLog from "./availability-log";
 import { isDeepStrictEqual } from "util";
 import { minimumAgeForProducts } from "./vaccines";
+import { dogstatsd } from "./datadog";
 
 // When locations are queried in batches (e.g. when iterating over extremely
 // large result sets), query this many records at a time.
@@ -81,7 +82,10 @@ function selectSqlPoint(column: string): string {
  * Create a provider location.
  * @param data ProviderLocation-like object with data to insert
  */
-export async function createLocation(data: any): Promise<ProviderLocation> {
+export async function createLocation(
+  data: any,
+  { source = "" } = {}
+): Promise<ProviderLocation> {
   data = validateLocationInput(data, true);
 
   const now = new Date();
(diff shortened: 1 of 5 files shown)


server/src/db.ts
```diff
@@ -158,7 +163,7 @@ export async function addExternalIds(
 export async function updateLocation(
   location: ProviderLocation,
   data: any,
-  { mergeSubfields = true } = {}
+  { mergeSubfields = true, source = "" } = {}
```
astonm, line 166: Seems useful to set a default here?

server/src/middleware.ts
```diff
@@ -39,3 +41,28 @@ export function versionedMiddleware(
...
+    ...otherOptions,
+    verify(
+      request: AppRequest,
+      response: Response,
+      buffer: Buffer,
+      encoding: string
+    ) {
+      request.bodyByteLength = buffer.byteLength || buffer.length;
+      if (verify) {
+        verify(request, response, buffer, encoding);
```
astonm, line 64: It's a little confusing here that this function is named `verify` and then inside calls a function `verify that is not itself. It might be less confusing to just make the call here as `options.verify` instead to differentiate the two. Alternately, it might be clearer to writ line 56 as `verify: (...) => { ...`?

astonm (approved):
👍🏽

server/src/db.ts
Mr0grog: Hmmmm, like what? This is the source from the loader (maybe I should call it `loaderSource`?), so I’m not sure how we’d pick any reasonable default, other than "null" or "?" or something.

astonm: Maybe `"unknown"`? Just trying to avoid `source:` showing up somewhere accidentally.

Mr0grog: Done.

astonm (approved):
:shipit:

PR #930, opened 2022-09-13
Make Render the production deployment! 🎉

### LINKED ISSUE #683: Explore running UNIVAF on Render

Mr0grog:
This is the first step of doing #779 for real (there’s plenty of cleanup to follow), and resolves #683. It flips various not-quite-production configuration values to production for Render, and makes it possible to point CloudFront at Render instead of our load balancer in AWS.

**Notes:**
- In #779, I originally pointed the DNS directly at Render. I later realized we should probably keep CloudFront since Render provides built-in DDOS protection (via Cloudflare), but not everyday caching, which is mainly what we use CloudFront for.
    - I set this up right now to support pointing to either our load balancer in AWS or to Render depending on the `api_remote_domain_name` configuration variable in Terraform Cloud. That way, we can easily flip it back if needed.
    - That means this won’t start pointing to Render until I manually *set* the variable in Terraform Cloud.
- In #779, I added `getmyvax.org` and `www.getmyvax.org` as domains to Render, but I did not do that here, because I *think* the CloudFront requests will come through with a different `Host` header. Adding the domains causes Render to try and verify via DNS, so we also need to temporarily point DNS directly at Render if we do that, then flip it back to pointing at CloudFront. So I’m hoping to avoid needing to do that.


Mr0grog:
This is did not work. Going to have to spend some time messing with CloudFront tomorrow to understand exactly what went wrong here..

PR #931, opened 2022-09-13
If there's no route, tag metrics with URL path

Mr0grog:
While reviewing the improved metrics from #915 earlier today, I noticed that a lot of external requests have no route, which turns out to be because the static file server doesn't count as a route (or routes). For example, `/docs/` will wind up with no route in our metrics.

Screen Shot 2022-09-12 at 8 33 58 PM

I tried to remedy the issue somewhat by filling in the requested URL (minus the querystring) when there’s no matched route, but realized that needs some guardrails — every tag value (and combination of tag values) counts as a metric we have to pay for, so if we allow any URL here, our DataDog bill is likely to blow up. I used status code < 300 (success & not a redirect) or status >= 500 (an error on our part, not the client’s) as a signal that the URL is likely legit and worth recording, although it’s certainly possible there might be a way someone could cause a set of random, non-legit URLs to cough up 5xx errors. I don’t think that’s happening or likely, though.

Notably, I excluded 3xx responses from the above, but they might also be worth recording? That said, I think the only real redirect we have is an explicit one (`/` → `/docs/`), and it already gets recorded (and will continue to) as a route.

### DIFF
--- a/server/src/datadog.ts
+++ b/server/src/datadog.ts
@@ -35,8 +35,17 @@ function replacePipeChar(text: string | RegExp): string {
   return text && text.replace(REGEX_PIPE, DELIMITER);
 }
 
-function getRoute(req: MonitoredRequest): string {
-  const routePath = req.route && req.route.path ? req.route.path : "";
+function getRoute(res: Response): string {
+  let routePath = res.req.route?.path || "";
+
+  // If a middleware responds (e.g. to serve static files), there won’t be a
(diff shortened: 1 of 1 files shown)


Mr0grog:
Anyway, this will start recording routes like `/docs/` and `/docs/openapi.yaml`, but not `/favicon.ico`.

Mr0grog:
Going ahead and landing this.

PR #934, opened 2022-09-13
Point `render.getmyvax.org` at CloudFront

Mr0grog:
This is the follow-up to #933. Render has verified the hostname and issued itself an SSL certificate, so now we can move DNS to point at CloudFront instead.


Mr0grog:
OK, this seems to be working, although it sure makes me feel sad to see both Cloudflare and CloudFront cache headers in the responses. 🙄

PR #935, opened 2022-09-13
Add public cache directive for Cloudflare

Mr0grog:
Let's see if making the `Cache-Control` directive explicitly allow public caching gets Render's built-in Cloudflare support to cache beyond just DDOS protection. If so, we can drop CloudFront.


Mr0grog:
Unfortunately, does not seem to be working. Probably because the built-in Cloudflare stuff is only DDOS protection.

PR #938, opened 2022-09-13
Fuzzy match new bivalent vaccines properly

Mr0grog:
This is kind of ugly, but attempts to handle new bivalent vaccines in a meaningful way when fuzzy matching (this mainly affects PrepMod, where we were conflating bivalent vaccines with normal ones).

Caught by the Anchorage i-Team.


Mr0grog:
Going ahead and shipping to solve things for Alaska folks.

PR #939, opened 2022-09-14
Followup/cleanup after switching production to Render

Mr0grog:
This cleans up some immediate leftovers from the DNS switch that made Render our production deployment. It:
- Removes a test DNS record and CloudFront distribution
- Removes the schedule for the daily data snapshot task, since we don’t want snapshots from AWS to overwrite or interfere with what Render is writing to S3.
- Makes it possible to disable loaders, but does not yet disable them (except for one, which was already effectively disabled via use of a wonky schedule).

These are needed for tonight. Tomorrow (Wednesday) or Thursday I’ll disable the loaders in AWS, and then start to actually remove AWS resources from existence!


Mr0grog:
Looks better in the Terraform plan this time! :shipit:

PR #940, opened 2022-09-15
Update docs for Render

### LINKED ISSUE #909: Add instructions for copying DB to runbook
When I was first setting up Render, I copied the contents of the production database so we had the same IDs and similar historical data. As part of doing so, I realized it was neither necessarily nor easily feasible to copy the `availability_log` table, so I didn’t (well, I didn’t think about this first, and the DB dump took so long I canceled it thinking something must have been wrong). When I created the new DB in Render and turned on all the other services, I had all kinds of errors because the API was trying to write to `availability_log`, which didn’t exist!

Yesterday I created a new, smaller DB to test with on Render (see #908), and made the same mistake again. I should add something to the runbook about copying the database, since you need schema+data for some of it, and only the schema for some of it. Basically:

```sh
# Dump schema + data for everything except availability_log
# (The log has too much data, and is not really important to copy most of time.)
pg_dump -d "${DB_URL}" \
    --no-owner \
    --exclude-table availability_log \
    > /data/univaf_dump.sql

# Add schema for availability_log
pg_dump -d "${DB_URL}" \
    --no-owner \
    --schema-only \
    --table availability_log \
    >> /data/univaf_dump.sql
```

Mr0grog:
Updates the README and runbook docs to match our new deployment environment in Render. Part of #683/#779. Also fixes #909.


astonm (approved):
Thanks for all the great updates here. Super thorough job, and I think it all reflects reality as I understand it.

One small tweak, a lot of the info you have around downgrading the database would be valuable for folks doing any sort of work with the database (including moving to a new provider, or even something as simple as pulling down some production data to work locally). Maybe consider reorganizing to make those steps more tactical ("how to dump the db") and then write up the "how to downgrade" section as pointers to those tactical steps?

Mr0grog:
Yeah, I was thinking about that a little bit when I wrote it, but decided to stay lazy. Thanks for the push break it up and clean it up. Will do.

PR #942, opened 2022-09-15
9/22: Stop building and deploying Docker images to ECR

Mr0grog:
Part of #683/#779.

Since our Render deployment is not Docker-based and we are shutting off things in AWS ECS, it no longer makes sense to build and publish Docker images. This does not remove our Docker files, both so it’s not too hard to spin things back up again if we have to switch back to ECS, and since they are referenced in our README as part of one kind of local development workflow (even though it’s one I don’t use anymore). After finishing the AWS teardown, I want to revisit local development and make some clearer decisions about Docker (also #882).

I’m going to let this sit for a little while before merging (at least until tomorrow — September 16 — or maybe next week), but just setting this up for easy review and merging then.


Mr0grog:
Thinking about order of operations here, I’m not going to land this until late next week, around when I plan to delete the database. As long as we are still publishing images, reconstructing the system is just a matter of putting Terraform configurations back.

Issue #912, opened 2022-09-08
Consider using DataDog HTTP API for metrics


Mr0grog:
We currently send metrics to Datadog using DogStatsD, which requires having the Datadog agent running. This has worked pretty well for us on AWS/ECS, but is less than ideal on Render — you can’t have sidecar containers in Render (so the agent can’t run alongside the app). Instead, the recommended solution is to run the agent as its own service. Unfortunately, running it as its own service means:
- The `host` tag is always wrong (it refers to the service the agent is running in, not whoever is sending metrics).
- The agent sends normal machine stats (CPU usage, memory, etc.) for itself, which is entirely pointless in this situation.
- The agent costs $18 on Datadog to be a host.

One alternative is to use Docker on Render and embed the agent as a background process in the app’s Docker container. That’s generally considered to be a poor pattern for containers and adds a lot of complexity (and we are switching to Render to *reduce* complexity!).

The better answer here is probably to use Datadog’s HTTP API to send metrics. A quick look around NPM makes [datadog-metrics](https://www.npmjs.com/package/datadog-metrics) look like the best option here — it gives us a very similar interface to the hot-shots package we are currently using for StatsD, and buffers & batches metric submission so we don’t have a lot of overhead from API requests. There’s also [dogapi](https://www.npmjs.com/package/dogapi), but the API is not as nice and it doesn’t do any buffering.

One issue that has to be solved is dev/test support. If you create an instance of the datadog-metrics client with no valid connection info, it throws errors when you send. (This works fine with hot-shots because StatsD is UDP-based, so it just fires messages off into the void and doesn’t care if nobody receives them.) We’d need to make a mock version or some kind of wrapper to use when a Datadog API key is not configured.

Mr0grog:
It would still be nice to use the same mechanism for Datadog in both the server and loader, but we rolled back the switch to Render so it's no really that critical. We've also shut down the entire production service, so I don't think this is worth addressing anymore. It is at best a theoretical nice-to-have.

PR #943, opened 2022-09-15
Use DataDog API instead of Agent in Render

Mr0grog:
This is one approach to #912. I started with the idea that “hot-shots and datadog-metrics have basically the same API, so we can just choose which agent and expose it! Minimal change!” But of course nothing ever quite works out that simply. (It pans out for most callsites, but configuration so it mostly matches DogStatsD’s behavior and tests are both messy.) I stopped here to keep the changes (and my time investment) reasonably small.

This works well enough to probably be worth trying, but some other possible approaches:

- Instead of swapping between the two, replace some callsites with the API and leave others pointing to the StatsD/hot-shots implementation.

- Abstract this into a “metrics” module (or some similar name), which chooses where and how to send (maybe even other providers instead of DD in the future? 🤷). This probably makes sense to do regardless and a future change.

- Separate the middleware from the DataDog client configuration. The way they are currently tied together makes the tests messier than they need to be (you can’t mock the DataDog client, or only sort-of can). Doing the above abstraction probably necessarily implies making this change, too. I didn’t do it here because I was trying to keep the scope smaller.


server/test/datadog.test.ts
```diff
@@ -36,19 +45,38 @@ describe("datadog middleware", () => {
...
+      expect(responseCounts[0]).toHaveProperty(
+        "tags",
+        expect.arrayContaining(["response_code:200"])
+      );
+
+      const responseTimes = buffer.filter(
+        (m) => m.key === "node.express.router.response_time"
+      );
+      expect(responseTimes).toHaveLength(1);
+    }
```
Mr0grog, line 79: This mess is dumb. The right move would be to have test mocks on the `dogMetrics` object so we can just check which functions were called, but since the middle we’re testing and the `dogMetrics` object are in the same module, they are tightly coupled through local references and the middleware will never see the mock (because we can only really mock *between* modules).

I can resolve this by separating the middleware out of `datadot.ts`, but that would imply even more architectural changes, and it just didn’t feel worthwhile in *this* PR, which I’m viewing as more running an experiment than changing implementations.

Mr0grog: I was trying to use a Jest mock, but for some reason I was hung up on mocking the whole `dogMetrics` object as one (which you could only do by mocking the module, since it’s an export, and the problem is that if you mock the module, the middleware will still see the real thing instead of the mock). Coming back with fresh eyes, we could just mock each method on the `dogMetrics` object individually:

```ts
jest.spyOn(dogMetrics, "increment");
jest.spyOn(dogMetrics, "gauge");
jest.spyOn(dogMetrics, "histogram");

afterEach(() => {
  jest.clearAllMocks();
});

// ...tests...
```

Mr0grog:
Putting this on hold for now. Even though we’re in a weird spot, where the loader uses the Datadog API but the server uses statsD and the Datadog Agent, there are some issues where posts to the API occasionally fail and I’m not confident in switching the server (questions to answer: are retries happening? how often is the agent dropping things for similar reasons?)

Mr0grog:
In the intervening time:
- We rolled back the switch to Render, and this became not as big a deal. (Although different parts of the codebase communicating with Datadog in different was is slightly confusing and lame.)
- Lots of surrounding code has changed, and this needs extensive rebasing and cleanup.
- A bunch of the funky hacks I did here have been solved upstream in the `datadog-metrics` package (yay!).

So some significant rework is needed here. But we've also shut down the production version of this project, and this work is probably not worthwhile in the end. I’m closing this PR.

PR #944, opened 2022-09-15
Add more detailed tracing for Sentry

Mr0grog:
This uses Sentry's tracing support to add some more spans around various operations (mostly in the update path) so we get more detailed traces. There's a lot "seems like it should be built in" utility tooling I had to write here, which was a bit disappointing.

It’s messy, and makes some of our code more complex (returning early is hard, as you can see in the “updateAvailability” route handler), and I’m not 100% sure this was totally worthwhile. Gonna let it sit for a while. May throw it out. 🤷 

Before:
Screen Shot 2022-09-15 at 12 25 16 PM

After:
Screen Shot 2022-09-15 at 12 25 35 PM

### DIFF
--- a/server/src/db.ts
+++ b/server/src/db.ts
@@ -19,7 +19,7 @@ import * as availabilityLog from "./availability-log";
 import { isDeepStrictEqual } from "util";
 import { minimumAgeForProducts } from "./vaccines";
 import { dogstatsd } from "./datadog";
-import { callWithSpan } from "./tracing";
+import { callWithSpan, finishSpan, startSpan } from "./tracing";
 
 // When locations are queried in batches (e.g. when iterating over extremely
 // large result sets), query this many records at a time.
@@ -87,7 +87,10 @@ export async function createLocation(
   data: any,
   { source = "unknown" } = {}
 ): Promise<ProviderLocation> {
-  data = callWithSpan(validateLocationInput, data, true);
+  const span = startSpan({ op: "validateLocationInput " });
+  data = validateLocationInput(data, true);
+  finishSpan(span);
+  // data = callWithSpan(validateLocationInput, data, true);
 
   const now = new Date();
   const sqlData: { [index: string]: string } = {
@@ -166,7 +169,10 @@ export async function updateLocation(
   data: any,
   { mergeSubfields = true, source = "unknown" } = {}
 ): Promise<void> {
-  data = callWithSpan(validateLocationInput, data);
+  const span = startSpan({ op: "validateLocationInput " });
+  data = validateLocationInput(data);
+  finishSpan(span);
+  // data = callWithSpan(validateLocationInput, data);
 
   const sqlData: any = { updated_at: new Date() };
 
@@ -585,7 +591,10 @@ export async function updateAvailability(
   id: string,
   data: AvailabilityInput
 ): Promise<{ action: string; locationId: string }> {
-  data = callWithSpan(validateAvailabilityInput, data);
+  const span = startSpan({ op: "validateAvailabilityInput " });
+  data = validateAvailabilityInput(data);
+  finishSpan(span);
+  // data = callWithSpan(validateAvailabilityInput, data);
   const {
     source,
     available = Availability.UNKNOWN,
--- a/server/src/db.ts
+++ b/server/src/db.ts
@@ -19,6 +19,7 @@ import * as availabilityLog from "./availability-log";
 import { isDeepStrictEqual } from "util";
 import { minimumAgeForProducts } from "./vaccines";
 import { dogstatsd } from "./datadog";
+import { callWithSpan } from "./tracing";
 
 // When locations are queried in batches (e.g. when iterating over extremely
 // large result sets), query this many records at a time.
@@ -86,7 +87,7 @@ export async function createLocation(
   data: any,
   { source = "unknown" } = {}
 ): Promise<ProviderLocation> {
-  data = validateLocationInput(data, true);
+  data = callWithSpan(validateLocationInput, data, true);
 
   const now = new Date();
   const sqlData: { [index: string]: string } = {
@@ -165,7 +166,8 @@ export async function updateLocation(
   data: any,
   { mergeSubfields = true, source = "unknown" } = {}
 ): Promise<void> {
-  data = validateLocationInput(data);
+  data = callWithSpan(validateLocationInput, data);
+
   const sqlData: any = { updated_at: new Date() };
 
   for (let [key, value] of Object.entries(data)) {
@@ -583,7 +585,7 @@ export async function updateAvailability(
   id: string,
   data: AvailabilityInput
 ): Promise<{ action: string; locationId: string }> {
-  data = validateAvailabilityInput(data);
+  data = callWithSpan(validateAvailabilityInput, data);
   const {
     source,
     available = Availability.UNKNOWN,
--- a/server/src/db.ts
+++ b/server/src/db.ts
@@ -240,12 +240,10 @@ export async function listLocations({
     }
 
     const locationIds = result.rows.map((r: any) => r.id);
-    const externalIds = await getExternalIdsByLocation(locationIds);
-    const availabilities = await getCurrentAvailabilityByLocation(
-      locationIds,
-      includePrivate,
-      sources
-    );
+    const [externalIds, availabilities] = await Promise.all([
+      getExternalIdsByLocation(locationIds),
+      getCurrentAvailabilityByLocation(locationIds, includePrivate, sources),
+    ]);
 
     return withSpan("joinRelatedRecords", () =>
       result.rows.map((row: any) => {
--- a/server/src/db.ts
+++ b/server/src/db.ts
@@ -19,7 +19,7 @@ import * as availabilityLog from "./availability-log";
 import { isDeepStrictEqual } from "util";
 import { minimumAgeForProducts } from "./vaccines";
 import { dogstatsd } from "./datadog";
-import { callWithSpan, finishSpan, startSpan } from "./tracing";
+import { finishSpan, startSpan } from "./tracing";
 
 // When locations are queried in batches (e.g. when iterating over extremely
 // large result sets), query this many records at a time.
--- a/server/src/db.ts
+++ b/server/src/db.ts
@@ -88,8 +88,11 @@ export async function createLocation(
   { source = "unknown" } = {}
 ): Promise<ProviderLocation> {
   const span = startSpan({ op: "validateLocationInput " });
-  data = validateLocationInput(data, true);
-  finishSpan(span);
+  try {
+    data = validateLocationInput(data, true);
+  } finally {
+    finishSpan(span);
+  }
   // data = callWithSpan(validateLocationInput, data, true);
 
   const now = new Date();
@@ -170,8 +173,11 @@ export async function updateLocation(
   { mergeSubfields = true, source = "unknown" } = {}
 ): Promise<void> {
   const span = startSpan({ op: "validateLocationInput " });
-  data = validateLocationInput(data);
-  finishSpan(span);
+  try {
+    data = validateLocationInput(data);
+  } finally {
+    finishSpan(span);
+  }
   // data = callWithSpan(validateLocationInput, data);
 
   const sqlData: any = { updated_at: new Date() };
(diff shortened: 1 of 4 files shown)


server/src/db.ts
```diff
@@ -86,7 +87,7 @@ export async function createLocation(
   data: any,
   { source = "unknown" } = {}
 ): Promise<ProviderLocation> {
-  data = validateLocationInput(data, true);
+  data = callWithSpan(validateLocationInput, data, true);
```
Mr0grog, line 90: Not sure whether I should have just used `wrapWithSpan` on all the validation function instead, so this *just happens* instead of having to explicitly use `callWithSpan` here.

server/src/tracing.ts
```diff
@@ -0,0 +1,142 @@
...
+    scope = Sentry.getCurrentHub().getScope();
+    parentSpan = scope.getSpan();
+  }
+
+  let newSpan: Span;
+  if (parentSpan) {
+    newSpan = parentSpan.startChild(spanOptions);
+  } else {
+    newSpan = Sentry.startTransaction(spanOptions as any);
+  }
```
Mr0grog, line 40: Ugh, just realized that, because `startTransaction` is an “extension” method (i.e. it doesn’t come from sentry/core or sentry/node), it doesn’t exist if `Sentry.init()` hasn’t been called, which is the case in some tests. That causes this to silently return `undefined` in those tests, which is causing them to fail. ugh.

server/src/tracing.ts
```diff
@@ -0,0 +1,142 @@
...
+ * @example
+ * withSpan({ op: "validateData" }, (span) => {
+ *   doSomeDataValidation();
+ * });
+ *
+ * let data = { some: "data" };
+ * const id = await withSpan({ op: "saveData" }, async (span) => {
+ *   const id = await saveData(data);
+ *   await updateSomeRelatedRecord(id, otherData);
+ * });
```
astonm, line 88: This async function should return the `id` value, no?

server/src/tracing.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  let callbackResult;
+  try {
+    callbackResult = callback(span);
+  } catch (error) {
+    finishSpan(span);
+    throw error;
+  }
+
+  if ("then" in callbackResult && "finally" in callbackResult) {
```
astonm, line 104: Is this the idiomatic way to test whether something is a promise?

astonm (commented):
I like the broad ideas in this PR (improve tracing, and add some utility functions for managing spans while we're at it), but something rubs me the wrong way with the `withSpan`, `wrapWithSpan`,`callWithSpan` API.

My first instinct is, I think it might be better to simplify it down to just `withSpan` and skip the other functions. They're a little bit subtle in their behavior–it took me a few reads of your docs to tell why/when I'd use them–and I'm not sure most cases the abstraction will save many lines of code. Also, as you sort of hint above, with multiple ways to do the same thing, it's not clear whether it's "right" to make the span wrapping obvious near the call site vs. abstracting it within the called function.

All that said, since Javascript doesn't have Python-style context managers that act like real blocks of code and respect stuff like `return`ing to the outer function's caller, maybe it's a mistake to do `withSpan`, too. Dumb question, but how would things look if everywhere we might initiate a `withSpan` instead we just explicitly call `startSpan` and `finishSpan`? I imagine exception handling would be an issue, but otherwise the code would be easier for someone to get their head around.

server/src/tracing.ts
Mr0grog: Argh, yes

server/src/tracing.ts
Mr0grog: Most things in the language just check for `then` (e.g. you can `await` any object with a `then(successCallback, errorCallback)`), so yes, more-or-less. I added the `finally` check since we are using finally here.

IIRC there’s some module we have that uses Bluebird instead of native promises (although maybe I got rid of it; I remember doing some cleanup at one point), and this kind of check instead of `x instanceof Promise` will work with non-native promises.

Mr0grog:
> how would things look if everywhere we might initiate a `withSpan` instead we just explicitly call `startSpan` and `finishSpan`? I imagine exception handling would be an issue, but otherwise the code would be easier for someone to get their head around.

Yeah, exception handling is *the* problem I was trying to handle here. Otherwise it’s a little noisy when there are a bunch of early returns (e.g. in the span I called `updateAvailability`).

That said, I guess a plain old `try...finally` covers both early returns and exceptions:

```js
const span = startSpan({ op: "thingThatMightReturnEarly" });
try {
  // do some stuff that might throw
  if (whatever) {
    return;
  }
  // do some other stuff that might throw
} finally {
  // runs for early returns and exceptions
  finishSpan(span);
}
```

So a little more boilerplate, but doesn’t need any fancy footwork to make early returns nice.

Mr0grog:
I guess another option here is to stop using `sendError()` and instead do `throw ValueError(message)`. (Should probably update the top-level error handler to [not log stack traces on 4xx errors](https://github.com/usdigitalresponse/univaf/blob/6dc7c50e94d65ae022641f63c51181e850547e3f/server/src/app.ts#L176-L177) in that case.)

Mr0grog:
In [v7.19.0](https://github.com/getsentry/sentry-javascript/releases/tag/7.19.0), Sentry added an OpenTelemetry integration. It might be interesting to see whether that is a good/better approach (debatable; it’s new and experimental, regardless of how mature OpenTelemetry might be).

- Release notes: https://github.com/getsentry/sentry-javascript/releases/tag/7.19.0
- Docs/source: https://github.com/getsentry/sentry-javascript/tree/966098be43dbe4d81b1250be559f4cf5010cfa1b/packages/opentelemetry-node

Mr0grog:
OK, so I tried this out without the `withSpan`/`callWithSpan`/`wrapWithSpan` mess, which resulted in me learning more frustrating Sentry SDK stuff: if you don’t explicitly finish a span, it will get dropped.

I added a fancy helper to try and address that by automatically closing spans when the transaction they are attached to closes (and nicely setting their status), but then realized that still leaves some edge cases that you need `try...finally` for (I did not yet add those in, so they aren’t in the diff). If you can rely on an exception never getting caught before closing the transaction, then that’s no big deal (nothing will come after the span in question). In the cases where I added spans that happens to be true, but it doesn’t really seem like a good practice to get in the habit of.

Adding the `try...finally` stuff so we’re doing it right feels like a step backwards for the places where we are just wrapping a function call (e.g. `callWithSpan(validateLocationInput, arg1, arg2, ...)`. It’s better, sorta, in `update()` endpoint where we have all the messy early returns. But not by a lot (OTOH, I guess this one place it’s safe not to use `finally` since the lifetime of the function *is* essentially the lifetime of the transaction).

server/src/tracing.ts
```diff
@@ -0,0 +1,187 @@
...
+  } else if (newSpan.transaction && newSpan.transaction !== newSpan) {
+    // FIXME: newer Sentry has an event for this: "finishTransaction" emitted
+    // on the hub's client:
+    // - Code: https://github.com/getsentry/sentry-javascript/blob/ba99e7cdf725725e5a1b99e9d814353dbb3ae2b6/packages/core/src/tracing/transaction.ts#L144-L147
+    // - Feature: https://github.com/getsentry/sentry-javascript/issues/7262
+    const _finishTransaction = newSpan.transaction.finish;
+    newSpan.transaction.finish = function finish(...args) {
+      cancelSpan(newSpan, "cancelled", "did_not_finish");
+      return _finishTransaction.call(this, ...args);
+    };
```
Mr0grog, line 71: More dumb lessons: the transaction event is on the client, which belongs to the hub. But the hub is stored on the transaction as a `_hub` property that you cannot get publicly!

So we can use the event if we grab a private property that might change, or we can keep this monkey-patching mess, which in practical terms is actually more stable. Sigh.

Mr0grog:
So, I think this is as good as it’s reasonably going to get. I don’t like it. This gets us:
- *Slightly* more detailed performance monitoring that I would like but which isn’t critical (e.g. I wanted to understand how much time we’re spending on validation), especially these days.
- Much more complex code and maintenance burden.
- Still some messy edge cases (reading up more, it seems like there are a lot of ways to get stuff on the wrong transactions or parent spans; Sentry uses Node.js domains, which don’t work great and are deprecated in favor of async hooks; OpenTelemetry uses async hooks but its integration w/ Sentry is still experimental, and the main Node.js OT package is also experimental itself, plus setup is complex).

I may actually just close this. Going to let it sit for a bit, but I think I may just close this. 😒

server/src/tracing.ts
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// Monkey-patch Transaction to add an event to notify listeners (in our case,
+// child spans) when the transaction is finishing. This is super hacky.
+//
+// The most recent release of the Sentry SDK has a `finishTransaction` event
+// on the hub's client, but there are a lot of guard clauses anywhere the client
+// gets used internally, and I'm not sure how reliable it is for this use case.
+// We want to wrap up before the transaction finishes, not when a client that
+// may-or-may-not exist depending on configuring is finishing a transaction.
```
astonm, line 26: Did you mean this?

```suggestion
// may-or-may-not exist depending on configuration is finishing a transaction.
```

server/src/tracing.ts
```diff
@@ -0,0 +1,202 @@
...
+Transaction.prototype.finish = function finish(
+  this: PatchedTransaction,
+  endTimestamp?: number
+) {
+  this._emitter?.emit("finish", this);
+  return _transactionFinish.call(this, endTimestamp);
+};
+
+// @ts-expect-error: _onFinish doesn't exist; we're adding it.
+Transaction.prototype._onFinish = function onFinish(
```
astonm, line 44: Any reason not to make this `onFinish` instead (without the underscore)?

server/src/tracing.ts
```diff
@@ -0,0 +1,202 @@
...
+ *
+ * More on why spans need canceling:
+ * https://github.com/getsentry/sentry-javascript/issues/4165#issuecomment-971424754
+ */
+export function startSpan(options: SpanOptions): Span {
+  let { parentSpan, timeout, ...spanOptions } = options;
+
+  let scope;
+  if (!parentSpan) {
+    scope = Sentry.getCurrentHub().getScope();
```
astonm, line 86: Given your comment above about `_hub` being nullable, are you sure `Sentry.getCurrentHub()` will always return an object that `getScope()` can be called on?

server/src/tracing.ts
```diff
@@ -0,0 +1,202 @@
...
+ */
+export function finishSpan(span: Span, timestamp?: number): void {
+  if (span.endTimestamp) return;
+
+  span.finish(timestamp);
+
+  const scope = Sentry.getCurrentHub().getScope();
+  if (scope.getSpan() === span) {
+    const parent = getParentSpan(span);
+    scope.setSpan(parent);
```
astonm, line 127: In the case where the span has no parent, what happens? Is it okay to set to e.g. a null span?

astonm (approved):
Lots of working and reworking here, so I can understand why you might still feel a bit unsettled. That said, I think where you landed is actually pretty good. A few small questions below, but mostly I think at this point the big risks are just the unknown unknowns building on top of Sentry stuff that clearly was not built with an eye towards extension this way.

Definitely sit on it for a bit, but I'm guessing the benefits of visibility could easily outweight the costs/risks here.

server/src/tracing.ts
Mr0grog: Mostly I was worried about colliding with some future feature. Then again, maybe it’s fine — if this method ever collided with an official feature, I’d be pretty darn surprised if was functionally different!

server/src/tracing.ts
Mr0grog: Yeah, this part is kinda complicated: `getCurrentHub()` always returns a hub (if there’s not one, it creates one). It’s the *transaction’s* (or the parent span’s transaction’s) `_hub` property that may or may not be set, because the transaction is given its hub at some weird time in its lifecycle, and I think it also depends on sampling rules and other settings. (Also the hub may or may not have a client, also depending on configuration! One of Sentry’s SDK devs told me they are both reliable if the transaction has been sampled, but then you have to check *that,* and who knows what future internal refactoring in the Sentry SDK could change that rule.)

(Also, `getCurrentHub()` returns a different hub for every async domain [at some point they’re going to upgrade to async hooks/storage] so Sentry can have concurrent scopes, like for two requests happening at the same time. So it matters that we use the right scope, since the parent span’s transaction’s scope might not be from the current async domain, so choosing when to get a scope via which method is a whole thing. 😬)

server/src/tracing.ts
Mr0grog: Yep — as long as nothing has gone *horribly* awry through misuse of the SDK somewhere, the only kind of span that doesn’t have a parent should be a transaction. When you end the transaction, you should remove it from the scope, which is what winds up happening here.

Mr0grog:
OK, I’m going ahead and merging, though I do worry a bit about the maintenance burden here.

PR #947, opened 2022-09-16
Fix a DataDog agent issue with ECS_FARGATE hack

Mr0grog:
Running the DataDog agent as a separate service (currently recommended by Render) comes with a bunch of caveats, including that it sends a whole raft of system metrics that are pointless, since nothing except the agent is actually run on the system it's gathering metrics for. Making the agent think it's on ECS Fargate using the `ECS_FARGATE` environment variable causes it not to send those. This *may* also remove a `host` from our DataDog account (saving some money), but I haven't seen for sure whether that's true yet.

### DIFF
--- a/render.yaml
+++ b/render.yaml
@@ -14,6 +14,14 @@ databases:
     ipAllowList: []
 
 services:
+  # Run the DataDog agent as a separate service, effectively making it a proxy
+  # for custom metrics sending from our other services.
+  #
+  # While this is recommended by Render, it's definitely not what the agent was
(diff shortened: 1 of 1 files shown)


Mr0grog:
Switching to the API instead of the agent entirely (see #943) would obviate this, but in the mean time, it’s probably useful. We’re actually already running with this setting right now, so this doesn’t change anything except putting it in the source and explaining it.

See also this discussion on Slack: https://usdigitalresponse.slack.com/archives/C01S4JWSMT7/p1661815283123649

PR #950, opened 2022-09-17
Delete API server & load balancer in AWS

Mr0grog:
Our primary DNS hasn’t been pointing at the API server for several days with no incident, so it’s about time to actually delete those resources.

This is a big one, and it’s the end of the day Friday. This one also involves a lot of cross-dependencies, so has a higher likelihood that normal for messy or broken applies. Leaving this in the queue for Monday.


Mr0grog:
Current plan: I’ll do this Monday, and wait until the second half of the week (Thursday?) to land #942 and delete the database. Then clean up the networks and logs and whatnot the week after, just before the end of the month.

astonm (approved):
👍🏽

PR #987, opened 2022-09-29
Stop trying to be so fancy with line breaks in PrepMod addresses

### LINKED ISSUE #929: Fix more complex PrepMod addresses
In #843, we punted on solving for locations where the line-splitting wasn’t unquestionably correct: https://github.com/usdigitalresponse/univaf/pull/843#issuecomment-1228050647

We’ve now seen a few of those, and have some Sentry issues caused by them:
- https://sentry.io/organizations/usdr/issues/3570497801: `"Quilcene School Multipurpose Room, 294715 US-101"`
- https://sentry.io/organizations/usdr/issues/3573194818: `"Quilcene Schools, Multipurpose Room, 294715 U.S. 101"` (in practice, this looks someone revised the first address 🙃)
- https://sentry.io/organizations/usdr/issues/3570442363: `"Chimacum School Gym, 91 West Valley Road"`
- https://sentry.io/organizations/usdr/issues/3573560728: `"Neighborhood House High Point | 6400 Sylvan Way Southwest"`

…so it’s probably time to solve this better.

There’s *kind* of a rule in the above examples: splitting when the next line would be `### [name] [street|way|road|whatever]`, although that wouldn’t help with the second one (which honestly would be valid to split or not split on the first comma).

That said, we still should have a more robust solution overall, since we know the above doesn’t cover all past cases we’ve seen (and therefore plenty more oddball ones we’ll see in the future).

From the original discussion:

> 1. Just let it send lots of events but ignore the ones that are OK (like these) in Sentry. Works, but not great because they’ll build up over time. (I also need to make sure these fingerprint differently and don’t all merge into one issue in Sentry, because then we can’t usefully ignore.)
>
> 2. **[This is what we did, and now need to go beyond]** These are both finding a line that is the suite/unit number. We could identify that pattern and always consider it OK (that is, move it from the `extraCleanLines` part to the `cleanLines` part). That’s nice, but is also just deferring this same problem until it’s something else that is fine, like `401 Coates Avenue Northwest, Yelm Middle School`, which is an actual (former) location we got from PrepMod.
>
> 3. Add a `corrections` module [like what we have for Albertsons](https://github.com/usdigitalresponse/univaf/blob/main/loader/src/sources/albertsons/corrections.js). I’m not sure this should be exactly the same, since PrepMod *appears* to have had some instances of addresses changing (BIG typos) in the past, so we maybe want to correct based on the address rather than based on the ID. So still some questions on the structure of this.
>
> 4. Add a list of addresses that are OK/correctable and which we don’t warn for. This would probably be in a separate module kind of like `corrections`.
>
> 5. A mix of 3 and 4 — 4 to solve this problem, but also 3 for things like complex instructions in the address (e.g. in the past we had a PrepMod location with the address `Inside Olympia Capital Mall, next to Buckle, closest parking in NW lot, near Shoe Dept. Encore, 625 Black Lake Boulevard Southwest`, and part of that address should be moved to the `description` field).
>
> 6. Something else?

I’ll add another now: just get rid of this whole [`cleanLines` vs. `extraCleanLines` thing](https://github.com/usdigitalresponse/univaf/blob/5878cc94a8f80dec4dcbe9f90110c4a450d52f36/loader/src/sources/prepmod/index.js#L173-L199), and not attempt to do any not-certainly-correct changes.

Mr0grog:
This is a competing proposal with #988.

Fixes #929 by ceasing to try to insert new line breaks in PrepMod addresses where they probably ought to be using all kinds of fussy logic. This is the simple approach and probably the smarter one. I’d be sad to lose all the nice address formatting, but I suspect it’s not worth the maintenance burden.

### DIFF
--- a/loader/src/sources/prepmod/index.js
+++ b/loader/src/sources/prepmod/index.js
@@ -143,20 +143,15 @@ function formatLocation(host, validTime, locationInfo) {
 const nonAddressLinePattern =
   /(^\s*|,\s+)[A-Za-z\s]+,\s+[A-Z]{2}\s*,?\s+(\d{5}(-\d{4})?|USA)\s*$/;
 
-// A lot of locations seem to have multiple address lines squished into one,
-// often using ", " or " | " or " / " as a separator. (" - " is common, too,
-// but is often used legitimately, e.g. in some road names.)
-const maybeAddressLineBreaks = /\s*,\s+|\s+[|/]\s+/g;
 // These are things that are *definitey* line breaks, as opposed to the above,
 // which are more fuzzy and should be surfaced for human review.
 const addressLineBreaks = new RegExp(
   [
-    // Actual line breaks
+    // Actual line breaks! :)
     String.raw`\s*\n\s*`,
-    // Things that might be line breaks if followed by a suite or unit number.
-    String.raw`(?:${maybeAddressLineBreaks.source})(?=(?:suite|ste\.?|unit)\s+#?\d+)`,
-    // " - " is a delimiter if followed by a suite or unit number, too.
-    String.raw`\s+-\s+(?=(?:suite|ste\.?|unit)\s+#?\d+)`,
+    // Things we've seen that appear seem safe to treat as line breaks.
+    // Spaces on both sides are important! These may not be breaks without them.
+    String.raw`\s+[|/]\s+`,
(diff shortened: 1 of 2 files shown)


Mr0grog:
Closed in favor of #988.

PR #988, opened 2022-09-29
PrepMod: Add line break in ", 123 somewhere rd"

### LINKED ISSUE #929: Fix more complex PrepMod addresses

Mr0grog:
This is a competing proposal with #987.

This fixes #929 by adding line breaks in situations like `", 123 Some Roadname Rd."` and `", PO Box 123"`. It also removes the Sentry warning, since this covers all the cases we've seen over an entire month.

I'm a little on the fence about this, since it involves so much added code for what feels like little benefit. Another reasonable approach here might be to just drop all the stuff here that is re-breaking lines entirely (the core part of the original problem was just to the redundant city/state/zip in the address lines, and this may have gotten out of hand). For example: 856c57c27125f7b96b1db150cdc4654a44a2cbe2

### DIFF
--- a/loader/src/sources/prepmod/index.js
+++ b/loader/src/sources/prepmod/index.js
@@ -6,7 +6,6 @@
  * contact at least one (and sometimes several) hosts in each state.
  */
 
-const { isDeepStrictEqual } = require("node:util");
 const { ApiClient } = require("../../api-client");
 const {
   EXTENSIONS,
--- a/loader/src/sources/prepmod/index.js
+++ b/loader/src/sources/prepmod/index.js
@@ -138,25 +138,69 @@ function formatLocation(host, validTime, locationInfo) {
   };
 }
 
+const streetTypes = [
+  "ave",
+  "avenue",
+  "dr",
+  "drive",
+  "rd",
+  "road",
+  "st",
+  "street",
+  "blvd",
+  "boulevard",
+  "ln",
+  "lane",
+  "cir",
+  "circle",
+  "ct",
+  "court",
+  "cor",
+  "corner",
+  "pl",
+  "place",
+  "plz",
+  "plaza",
+  "way",
+  "pkw?y",
+  "parkway",
+  "cswy",
+  "causeway",
+  "xing",
+  "crssng",
+  "crossing",
+  "sq",
+  "square",
+  "trl?",
+  "trail",
+].join("|");
+
 // Attempts to match "<city>, <state abbreviation> <zip>" so it can be removed
 // from the address lines.
 const nonAddressLinePattern =
   /(^\s*|,\s+)[A-Za-z\s]+,\s+[A-Z]{2}\s*,?\s+(\d{5}(-\d{4})?|USA)\s*$/;
 
 // A lot of locations seem to have multiple address lines squished into one,
-// often using ", " or " | " or " / " as a separator. (" - " is common, too,
-// but is often used legitimately, e.g. in some road names.)
-const maybeAddressLineBreaks = /\s*,\s+|\s+[|/]\s+/g;
+// often using ", " or " - ".
+const maybeAddressLineBreaks = /\s*,\s+|\s+-\s+/g;
 // These are things that are *definitey* line breaks, as opposed to the above,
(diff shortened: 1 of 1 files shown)


astonm (commented):
I see the appeal of the simpler solve in #987, but on the other hand, this doesn't look so unmaintainable to me.

It does make me wonder, is there a library that already does this sort of thing that we could pull in instead?

Mr0grog:
> is there a library that already does this sort of thing that we could pull in instead?

No, not that I was ever able to find and that quite fit our needs and/or data (witness also the absurdity of the `matchableAddress()` stuff in the `utils` module, which, in a happier and more organized world, would share some of its guts with this, e.g. the list of `streetTypes` — that's where I pulled them from, but making the two work together would involve more messy refactoring that I don't want to do right now). Anyway, I’m sure there there is plenty of similar-ish stuff in a variety of geocoding tools, but those are largely closed source and this slice of functionality isn't exposed.

Mr0grog:
This should probably get merged into `utils`, or all the address stuff from everywhere should go into one place, or *something* at some point, but I’m going ahead and merging this for now.

PR #1012, opened 2022-10-11
AWS again: Reinstate networks and bastion

Mr0grog:
This re-instates the old public/private network setup and bastion security group on AWS. I combined all network configuration into a single file this time (makes more sense to me) and tried to add more clearly explanatory comments. I'm not sure I got the network stuff 100% right (this is the part of AWS I have the most trouble with, TBH), but this matches what I understand from the previous comments.


Mr0grog:
Out-of-band 👍 from @astonm.

PR #1014, opened 2022-10-11
AWS again: reinstate database

Mr0grog:
Recreate the database on AWS. This uses smaller default instance and disk size than we used to have, and is in-line with what we were using on Render (which seemed to work just fine). Should save a little on costs.

### DIFF
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -28,6 +28,26 @@ variable "data_snapshot_s3_bucket" {
   default     = "univaf-data-snapshots"
 }
 
+variable "db_user" {
+  description = "The database user"
+  default     = "univaf"
+}
+
+variable "db_password" {
+  description = "The password for the database instance, filled via Terraform"
+  sensitive   = true
+}
+
+variable "db_instance" {
+  description = "The instance type for the DB. Reference: https://aws.amazon.com/rds/instance-types/"
+  default     = "db.t4g.small"
+}
+
+variable "db_size" {
+  description = "The storage size for the DB (in Gigabytes)"
+  default     = 30
+}
+
 # These AWS variables are present to clean up warnings in terraform
 variable "AWS_SECRET_ACCESS_KEY" {
   default = ""
--- a/terraform/db.tf
+++ b/terraform/db.tf
@@ -1,0 +1,23 @@
+
+module "db" {
+  source = "./modules/rds"
+
+  name     = "univaf-db"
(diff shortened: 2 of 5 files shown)


terraform/variables.tf
```diff
@@ -28,6 +28,26 @@ variable "data_snapshot_s3_bucket" {
...
+}
+
+variable "db_password" {
+  description = "The password for the database instance, filled via Terraform"
+  sensitive   = true
+}
+
+variable "db_instance" {
+  description = "The instance type for the DB. Reference: https://aws.amazon.com/rds/instance-types/"
+  default     = "db.t4g.small"
```
Mr0grog, line 43: Was previously `db.m5.large` (2 CPU / 8 GB RAM). This is 1 CPU (ARM) / 2 GB RAM, and matches what we have on Render (which seems to be fine). Costs ~$25/mo instead of ~$135/mo.

terraform/variables.tf
```diff
@@ -28,6 +28,26 @@ variable "data_snapshot_s3_bucket" {
...
+}
+
+variable "db_size" {
+  description = "The storage size for the DB (in Gigabytes)"
+  default     = 30
```
Mr0grog, line 48: Was previously `240`, which was a result of dealing with an long-ago outage related to clearing old logs. We always meant to turn this down, but never got to it.

In Render, we have a 48 GB disk and have not used more than 12 GB of it in the past week, so this ought to be plenty.

Mr0grog (commented):
All the code in `modules` is exactly the same as it used to be. The newer code has been rearranged slightly for readability and had the settings tuned down to match our actual needs (one advantage of the failed switch to Render was an opportunity to provision things more minimally).

terraform/variables.tf
Mr0grog: This is only the second time I’ve used Graviton instances for anything! They’re a little cheaper, and *should* perform just as well on less power. We can bump this up pretty easily if we see problems.

Mr0grog:
Note to self: need to update the SNS subscription that powers our Slack integration once this deploys!

astonm (approved):

Mr0grog:
> need to update the SNS subscription that powers our Slack integration once this deploys!

Enjoyably, the ARN for the alarm’s topic is based on its name, meaning this created a topic with the same ARN as the old one (we didn’t change the name), so the subscription that sends it to Slack just automatically picked it up! Didn’t actually have to do anything.

PR #1020, opened 2022-10-12
AWS again: reinstate daily data snapshot

Mr0grog:
**Depends on #1019, which should land and get tested first.**

This is the last component of a functioning deployment on AWS (besides DNS & CloudFront, which aren't strictly required for things to work).

Once this is in, we should wait for a day and make sure we feel good before switching DNS and CloudFront back over to AWS and turning off Render. This also gives us time to a little messy refactoring on the Terraform code before things are in production, and changes need to be made more carefully.

### DIFF
--- a/terraform/api.tf
+++ b/terraform/api.tf
@@ -28,19 +28,52 @@ resource "aws_route53_record" "api_www_domain_record" {
 }
 
 
-# These log groups and streams are associated with a tasks that no longer run,
-# but we are preserving the logs for a little while longer.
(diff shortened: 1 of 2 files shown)


Mr0grog:
Cleaned up and rebased.

Mr0grog:
NOTE: I’ve configured this to write to the old `univaf-render-test-data-snapshots` bucket for now, so it doesn’t conflict with Render while Render is still production.

PR #1021, opened 2022-10-12
Fix deprecation warnings for action-linter

Mr0grog:
We are getting deprecation warnings in our CI workflows. This cleans them up by switching to actionlint (the previous linter we used is no longer maintained, and they recommend actionlint instead).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,9 +152,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: cschleiden/actions-linter@v1
-        with:
-          workflows: '[".github/workflows/*.yaml", ".github/workflows/*.yml"]'
+      - name: Lint workflow files
+        run: |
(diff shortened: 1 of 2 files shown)


Mr0grog:
Hmmm, looks like we can’t fix the other warnings yet; the official AWS actions don’t have a new version with the right Node.js support yet: https://github.com/aws-actions/configure-aws-credentials/issues/489

.github/workflows/ci.yml
```diff
@@ -194,8 +200,8 @@ jobs:
               --format '{{ .VirtualSize }}' \
               "${IMAGE_NAME}"
           )
-          SIZE_MB=$(($SIZE / 1024 / 1024))
-          echo "Built image \`${IMAGE_NAME}\`: ${SIZE_MB} MB" >> $GITHUB_STEP_SUMMARY
+          SIZE_MB=$((SIZE / 1024 / 1024))
```
Mr0grog, line 203: TIL: `$` is not needed for variables inside math expressions in bash!

PR #1024, opened 2022-10-12
AWS again: point ECS domain to correct place

Mr0grog:
I wasn't paying attention in #1023 and pointed ecs.getmyvax.org to the Render CloudFront instead of the ECS CloudFront.

### DIFF
--- a/terraform/api.tf
+++ b/terraform/api.tf
@@ -307,14 +307,12 @@ resource "aws_cloudfront_distribution" "univaf_api_ecs" {
   enabled     = true
   price_class = "PriceClass_100" # North America
   aliases = [
-    var.domain_name,
-    "www.${var.domain_name}",
     "ecs.${var.domain_name}"
   ]
   http_version = "http2and3"
 
   origin {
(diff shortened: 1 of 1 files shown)


terraform/api.tf
```diff
@@ -301,11 +306,11 @@ resource "aws_cloudfront_distribution" "univaf_api_ecs" {
...
   price_class = "PriceClass_100" # North America
-  aliases = compact([
+  aliases = [
     var.domain_name,
-    length(aws_route53_record.api_www_domain_record) > 0 ? aws_route53_record.api_www_domain_record[0].fqdn : "",
-    length(aws_route53_record.api_ecs_domain_record) > 0 ? aws_route53_record.api_ecs_domain_record[0].fqdn : ""
-  ])
+    "www.${var.domain_name}",
+    "ecs.${var.domain_name}"
+  ]
```
Mr0grog, line 313: Now I realize why we didn’t use references before: they make an dependency loop in this situation.

PR #1030, opened 2022-10-14
Terraform Cleanup & Refactor

Mr0grog:
Before we make it production again, do some cleanup and refactoring to make our Terraform configuration files more approachable. Some things I’ve tried to do here:
- Add more explanatory comments.
- Move related resources together, rather than grouping by type of resource (e.g. the IAM role that you have to have in order to do things in the ECS cluster is now defined alongside the cluster, instead in a separate place).
- Name files more thematically (e.g. all the stuff related to the API server is named `api*.tf`).


Mr0grog:
Alright, I think this is good for now.

I thought about removing the `loader` module from Terraform entirely, but feel on the fence about it, so I left it for now. Originally, I thought I’d make it into a more generic `cron_job` module that the daily data snapshot job could also use, but the refactors here removed so much responsibility from it that I’m not sure it’s actually worth modularizing anymore. Here’s what that might look like, as an experiment: https://github.com/usdigitalresponse/univaf/commit/b19cb1f0b3358d81a0b7441e98ee5ae57a716c62

astonm (approved):
Took a look, and all of these factors seem good to me 👍🏽. I think it's easier to follow what's going on as well.

I like the change in https://github.com/usdigitalresponse/univaf/commit/b19cb1f0b3358d81a0b7441e98ee5ae57a716c62, but if you don't want to go that far here, fine with me.

Mr0grog:
Cool, if you think that extra change is good, I’ll add it this and merge.

Mr0grog:
Yeesh, ok, I’m landing this now.

PR #1037, opened 2022-10-17
Use more smaller API server instances

Mr0grog:
One lesson from rebuilding things on AWS is that we just don't use much memory in the API server. We can't reduce RAM without also cutting the CPU in half, but my working theory is that we can just double the instances and it should be fine. We'll follow the charts and see how this goes (honestly I think a baseline of more instances might make the whole thing more stable anyway).

### DIFF
--- a/render.yaml
+++ b/render.yaml
@@ -42,13 +42,13 @@ services:
   - name: API Server
     region: oregon
     type: web
-    plan: standard
+    plan: starter plus
     env: node
     repo: "https://github.com/usdigitalresponse/univaf.git"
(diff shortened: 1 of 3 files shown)


Mr0grog:
Well, that was clearly a bad idea: this slightly increased the error rate (still at a level that’s fine overall) and doubled our response latency (!), which is about the opposite of what I was hoping for.

Screen Shot 2022-10-18 at 9 20 27 AM

Screen Shot 2022-10-18 at 9 21 15 AM

PR #1075, opened 2022-10-25
Upgrade AWS SDK to v3.194.0 in /server

Mr0grog:
Upgrades the AWS SDK components in `/server` to v3.194.0, since Dependabot didn’t do it.

### DIFF
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -9,8 +9,8 @@
       "version": "0.1.0",
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


Mr0grog:
Weird, the combination that passed building here does not work for me locally, and the second commit fixes it for me locally but breaks here. :(

Mr0grog:
Ah, looks like I accidentally overrode my local config for this project and was building with Node 18 locally.

PR #1118, opened 2022-11-15
Upgrade Node.js to v18.12.1

Mr0grog:
Node 18 is now the LTS release, so it's probably time to upgrade to it from 16. There are some changes to Node.js types, but no new JS language features that warrant updating.


Mr0grog:
This will also fix the `@types/node` issues on #1084 (and other PRs it superseded). The problem there is that AWS has been using Node.js 18-specific types that don’t exist in Node 16. Since we are upgrading to Node 18, that’s no longer a problem (for us).

PR #1133, opened 2022-12-02
Share common code between loader and server, take 2

Mr0grog:
This is a re-do of #882. There have been enough changes in the intervening time that updating it was a little rough.

From the original PR:

> This is an attempt at making it possible for the loader and server to share code rather than having some stuff copy/pasted. The are 2 core goals:
> 1. Get rid of duplicated code that can get out of sync/produce subtle bugs.
> 2. Make getting started and working on the project a little easier.
> 
> This adds a third source code/project directory named `common` with a package named `univaf-common` in it, and also switches to using NPM workspaces to manage our dependencies instead of setting up each package independently > (running `npm install` in the project root installs common, loader, server, and ui all at once; this also allows us to import common code from `univaf-common` instead of relative file paths). The file structure looks like:
> 
> ```
> /
>   - package.json  (lists the workspaces - common, loader, server, ui)
>   - common/    (New)
>     - package.json
>     - tsconfig.json
>     - src/
>       - index.js
>       - jest.js
>       - states.json
>       - <whatever>.js
>   - loader/    (No consequential changes)
>     - package.json
>     - src/
>     - test/
>   - server/    (No consequential changes)
>     - package.json
>     - tsconfig.json
>     - src/
>     - test/
>   - ui/        (No consequential changes)
>     - package.json
>     - src/
> ```
> 
> The source in `common` is JavaScript (so the loader can use it directly), but has a `tsconfig.json` file so the server can understand it and do autocompletion, automatic typing and building (when you build the server, it will > build common for you if it’s out of date), etc. That said, we have some stuff (like the [vaccine constants](https://github.com/usdigitalresponse/univaf/blob/d50ccb052292c968b972d903305d111396492353/server/src/vaccines.ts#L1-L51)) > that really leverage TypeScript, and it would be a shame to lose that. So we may have to switch to making this package compiled, and make working on the loader a little more complicated.
> 
> I don’t want to move all the shared code possible in this PR. Instead, this should get a working framework in so that moving code to a shared location can be an incremental process. That said, some things that should still happen > in this PR or a PR that merges before this one:
> - Figure out how to handle sharing code that really leverages TypeScript (like the [vaccine constants](https://github.com/usdigitalresponse/univaf/blob/d50ccb052292c968b972d903305d111396492353/server/src/vaccines.ts#L1-L51))
>     - Making the `common` package TypeScript makes working on the loader awkward because you have to know to build `common` first.
>     - We could make the loader a TypeScript project (even if all the files stay as JS). As a bonus, we get type checking against our JSDoc comments (which will probably produce some errors from bad or unmaintained comments at > first).
>     - We could just have a build script in the loader that runs the build script in `common`.
> - Update docs.
> - Update our Dockerfiles and scripts/workflows that use them. They currently only include their subdirectly, but will now need to include the whole project. *OTOH: maybe we just drop the Docker stuff in #779, and then this isn’t a > problem.*
> - Set up tests for the `univaf-common` package.
> - Prettier and ESLint are still defined per-package, but could be moved to the root level.

## To-Do

- [x] Handle sharing code that really leverages TypeScript, e.g. the vaccine constants. *(I wound up just making the common module TypeScript. It makes things a little messier for making changes when working on the loader, but is still probably the best approach.)
- [x] Update Dockerfiles. *(There is now ONE Dockerfile with multiple targets, instead of a file in each directory.)*
- [x] Tests for the `univaf-common` package.
- [x] Move Prettier and ESLint to the root level?
- [x] Docs


astonm:
I haven't looked through the details yet, but I'm 👍🏽 on the general structure and approach here. Can you say more about your hesitation to make the loader a TypeScript package?

Mr0grog:
> Can you say more about your hesitation to make the loader a TypeScript package?

Changing all the `.js` files to `.ts`, that feels like a big project that’s not worthwhile. But if you’re asking about adding a `tsconfig.json` and an NPM `build` script, I started to do that and ran into some weird issues around tests. Ultimately I bailed on it because it didn’t seem strictly necessary to solve here, and mainly resulted in adding extra config files that were a bit special (compared to the other packages) because the code is mainly JS.

But I could definitely add them if you think it would make things better/more consistent. It would at least mean we’d just build the loader, instead of explicitly building `common` in order to use the loader.

Mr0grog:
(Side note: I tried turning on `checkJs` or putting `// @ts-check` directive at the top of some files and… there are a lot of not-quite accurate type comments that need cleanup. 😬 )

astonm:
Okay, sounds like it would be a really big lift to convert it. I was thinking it might be nice just to get everything consistent, and obviously fixing all the type errors would be great. But let's save it for a later code clean up, and go ahead 🚢 this one.

PR #1184, opened 2022-12-16
Add Docker layer caching in CI

Mr0grog:
Docker layers are big and include our dependencies, so we should probably be caching them in CI just like we do our dependencies for tests, linting, etc.

This makes use of the `docker/` GitHub actions instead of our own scripts (like we used to do) because they automatically set some of the configuration up for caching with GitHub Actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,7 +189,19 @@ jobs:
       - name: Cache Docker layers
         uses: actions/cache@v3
         with:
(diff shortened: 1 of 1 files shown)


Mr0grog:
I am surely missing some very obvious and basic thing here, but I sure couldn’t tell you what it is. I was hoping this could make the wait for builds significantly shorter, but that is not worth putting more time into at this point.

Mr0grog:
Nope, I give up for real now.

PR #1187, opened 2022-12-16
Only allow requests through CloudFront

### LINKED ISSUE #1181: Secure access to API load balancer
Right now, someone who knows the load balancer’s actual DNS can bypass CloudFront and make requests directly to the load balancer. AWS has a good guide on securing access to the origin load balancer here: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/restrict-access-to-load-balancer.html

And there are some examples (but less explanation) of how to do the same configuration in the Terraform docs here: https://registry.terraform.io/providers/hashicorp/aws/4.46.0/docs/resources/lb_listener_rule

We should also make sure the origin uses HTTPS only instead of HTTP. There’s not a whole lot we need to worry about for security in our case (everything over the wire here is generally public data), but this is still a good idea.

Mr0grog:
This locks down the API server's load balancer so that it only forwards requests to the actual service if it has a header with a secret value. CloudFront is already configured to set that header (see #1186), so this effectively allows only requests from CloudFront.

Fixes #1181.

### DIFF
--- a/terraform/api.tf
+++ b/terraform/api.tf
@@ -136,7 +136,7 @@ resource "aws_lb_listener_rule" "api_forward_if_secret_header" {
 
   # Add a condition requiring a secret header only if there's a secret to check.
   dynamic "condition" {
-    for_each = var.api_cloudfront_secret ? [1] : []
+    for_each = var.api_cloudfront_secret != "" ? [1] : []
     content {
       http_header {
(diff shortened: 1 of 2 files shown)


Mr0grog:
:shipit: 🤞

PR #1190, opened 2022-12-17
Actually disallow non-CloudFront traffic, take 2

Mr0grog:
This has been a messy afternoon. I thought I was good to go in #1189, but then realized I had accidentally started blocking the loaders from sending data to the API. This new approach should make sure loaders use HTTPS and the right domain for the load balancer, plus allow them through since they supply a valid API key.


Mr0grog:
Here we go again…

Issue #1180, opened 2022-12-13
Secure snapshots S3 bucket behind CloudFront


Mr0grog:
Since requests directly to S3 buckets can get expensive quickly, we should put our [“data snapshots” S3 bucket](https://github.com/usdigitalresponse/univaf/blob/2e0779b22ffec4cd449ce33870c8ee3f0ff22f99/terraform/s3.tf#L1-L28) behind CloudFront and disable public access. (This recommendation from @TylerHendrickson’s very helpful review of our AWS configuraiton.)

- [x] ~Make writes to S3 atomic. Since writes can be slow (e.g. each `availability_log` file takes just under 20 minutes to write), there’s a reasonable concern that CloudFront could cache a partially written file. Write to a key with a prefix, then move the data to a key without the prefix (e.g. write to `/in_progress/availability_log/availability_log-2022-12-12.ndjson.gz`, then move to `/availability_log/availability_log-2022-12-12.ndjson.gz`). See https://github.com/usdigitalresponse/univaf/blob/main/server/scripts/availability_dump.js~ *(Update: it turns out streamed, multipart uploads are already atomic, so we don’t have to do anything here.)*

- [x] Add a CloudFront distribution that reads from the bucket. It’ll need a domain name, so we should probably use `archives.getmyvax.org` or `snapshots.getmyvax.org`.

- [x] Update code for loading data in https://github.com/usdigitalresponse/appointment-data-insights. (PR: usdigitalresponse/appointment-data-insights#4)

- [ ] Make the bucket itself private instead of public.

Mr0grog:
TIL multipart uploads to S3 are atomic — it doesn’t assemble the pieces into a GETable object until all parts are uploaded (see https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html). Single operation PUTs are also atomic (see https://docs.aws.amazon.com/AmazonS3/latest/userguide/Welcome.html#ConsistencyModel)

So we actually don’t need to worry about item 1 (make writes atomic)!

PR #1264, opened 2023-01-31
Add CDN for archived data in S3

Mr0grog:
This adds a CloudFront distribution to serve data from our "data snaphsots" S3 bucket at `https://archives.getmyvax.org`. The goal here is mainly to prevent people from potentially driving up S3 costs by making requests we can't control or cache against the bucket. (Per @TylerHendrickson’s guidance.)

This is a first step for #1180. After we confirm this is deployed and working, we’ll need to revoke public read access from the S3 bucket.

### DIFF
--- a/terraform/api-domains.tf
+++ b/terraform/api-domains.tf
@@ -75,21 +75,6 @@ resource "aws_route53_record" "api_load_balancer_domain_record" {
   }
 }
 
-# # Serves archived, historical data from S3 via CloudFront.
-# resource "aws_route53_record" "data_snapshots_domain_record" {
-#   count = var.domain_name != "" ? 1 : 0
-
-#   zone_id = data.aws_route53_zone.domain_zone[0].zone_id
-#   name    = local.data_snapshots_domain
-#   type    = "A"
-
-#   alias {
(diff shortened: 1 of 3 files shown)


Mr0grog:
OK, I’ve implemented this both directly with the AWS provider and using a [module from Cloud Posse](https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn) that makes the setup significantly simpler. I think the simplicity is probably worthwhile, but there are some caveats worth logging here:

- It might be broken! It triggers one deprecation warning in Terraform which *should* be OK, but this issue suggests there may be bigger problems: https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/issues/217. It’s possible I’m just not using the problematic features, or maybe it’s since been fixed and nobody closed the issue. I guess we’ll see. 🤷

- It does so many things that properly understanding how to configure it correctly can be a bit confusing. See, for example, how many commits I made so I could see what Terraform’s plan output was and adjust (e.g. several settings determine whether alternate domain names are used, and the implications of that is not clearly documented anywhere). There is definitely some cost to this abstraction.

    (OTOH, I had forgotten to create an IPv6 DNS record in my manual version, and this module takes care of that. Good defaults, mostly.)

PR #1269, opened 2023-02-01
Clean up failed upload parts in S3

Mr0grog:
Today I learned that multipart S3 uploads can leave the parts behind if they fail, taking up space in S3 even though they don't show up as complete objects (thanks, @TylerHendrickson). Since all streaming uploads and large uploads done using the AWS libraries or CLI are multipart, this applies to all our uploads. This policy should help clean them up. (We've rarely seen failures, but this is still good practice.)

### DIFF
--- a/terraform/s3.tf
+++ b/terraform/s3.tf
@@ -51,7 +51,7 @@ resource "aws_s3_bucket_policy" "render_test_data_snapshots" {
   })
 }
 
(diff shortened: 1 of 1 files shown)


astonm (approved):
Nice, I didn't know 1) this happened and 2) you could get AWS to manage it for you. 💯

PR #1272, opened 2023-02-06
Take more care when writing logs to S3

Mr0grog:
At the end of last week, I discovered some bugs with how we write historical data to S3, and this should resolve them:
- Check for logs to be written before actually writing log files. Since we clear out old logs (they inflate the database unreasonably), not doing this previously led to us overwriting some valid log files with empty logs because we’d cleared the data for that date!
- Page through *all* objects in S3. The S3 APIs only list up to 1,000 objects, and we now have more than that many log files in the data snapshots bucket. This led to us overwriting recent log files because the script didn’t see that they already existed, and also led to the above issue where we overwrite valid logfiles with empty ones.

### DIFF
--- a/server/scripts/availability_dump.js
+++ b/server/scripts/availability_dump.js
@@ -70,6 +70,35 @@ function removeNullPropertiesStream() {
 }
 
 /**
+ * Get a set with the names of all the objects in an S3 bucket. This paginates
+ * through all results from the S3 API and may make many requests.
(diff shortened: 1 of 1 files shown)


astonm (approved):
Oof, good catches here. Apologies for not considering these issues back when we first put all this together.

Mr0grog:
C’est la vie! Thanks for taking a look.

PR #1295, opened 2023-02-13
Actually retry GraphQL errors

Mr0grog:
Back in #1266, I sorta-kinda added retry support for Hy-Vee. It reduced errors slightly, but they’re still annoyingly noisy in Slack and Sentry (the rate is not technically problematic, but [they happen enough to be distracting](https://usdr.sentry.io/issues/3515438646)). After more digging, I’m pretty sure what’s happening is that Hy-Vee’s GraphQL API returns errors (as long it was able to catch and format them as GraphQL errors) with a 200 HTTP status code (IIRC Washington’s uses meaningful status codes; the GraphQL spec does not say anything about HTTP status codes one way or the other). That means that our HTTP client’s built in retry functionality doesn’t work. (This conclusion lines up with the reduced errors — I haven’t seen any more *network* errors, e.g. `EAI_AGAIN`, `ECONNRESET` in this codepath since #1266.)

To handle this, I’ve added a `queryGraphQl(url, options)` helper that takes care of the retry logic (and added some more smarts to our exception objects, too). It also sets a few extra defaults that are relevant for GraphQL but not for RESTful requests. You can pass a `retryIf: (error) => boolean` option to determine whether a GraphQL error response should be retried.


astonm (approved):
Conceptually looks great. Hard to verify the nitty gritty behavior based on what's here, but the tests you added lend a good amount of confidence, imo.

Mr0grog:
Thanks for taking a look! Despite how much we see [the error in question](https://usdr.sentry.io/issues/3515438646/?project=5718483&query=is:unresolved) in production, I was only able to repro them once while working on this. So yeah, hard to verify concretely. We’ll see what happens. 😅

Mr0grog:
Circling back: no errors of this type in the past 48 hours! (Otherwise, in the past month, there has only been one day with none of these errors.) 🎉

PR #1307, opened 2023-02-21
Bump Jest to v29.4.3

Mr0grog:
*This is a redux of #1304 and #1306, which got incorrectly merged and reverted due to a failed experiment with GitHub’s new “merge queue” feature.*

Bumps [jest](https://github.com/facebook/jest/tree/HEAD/packages/jest) from 29.4.2 to 29.4.3.
- [Release notes](https://github.com/facebook/jest/releases)
- [Changelog](https://github.com/facebook/jest/blob/main/CHANGELOG.md)
- [Commits](https://github.com/facebook/jest/commits/v29.4.3/packages/jest)

---
updated-dependencies:
- dependency-name: jest dependency-type: direct:development update-type: version-update:semver-patch ...


Mr0grog:
This release of Jest fixed a long-standing bug (for multiple major versions) that we appear to have relied on: https://github.com/facebook/jest/issues/13925

Calling `jest.restoreAllMocks()` wasn’t previously restoring from manually mocked modules (among other things, but this is the one that matters for us), and now it does. Back in #1256, it looks like I added a `jest.restoreAllMocks()` call when using the using the new `replaceProperty()` feature, failing to notice there was *another* mock that we didn’t want to restore. Since there was a bug that caused that mock not to actually get restored, it all worked fine, and I blissfully merged the bad code in. Now that it’s been fixed upsteam, our tests started failing. I’ve fixed it here by separating out the mocks we want to clear vs. restore and calling the appropriate methods on each.

PR #1319, opened 2023-02-27
Support removing "Remote" domain name

Mr0grog:
We no longer have anything deployed in Render or other hosts, so make sure the Terraform configuration can handle removing the remote domain name variable.


Mr0grog:
I’ve also updated the variable on Terraform Cloud, so merging this will remove the domain name (`render.getmyvax.org`) and the CloudFront distribution.

PR #1322, opened 2023-03-01
Try running loaders on default subnets

Mr0grog:
We're getting pretty extreme Internet Gateway costs that appear to mostly be coming from the amount of data we pull in from outside parties in the loader. This is an experiment to see whether running the loaders in the default subnets avoids some of that cost.

(In #1321 I tried compressing the data we send from the loaders, but when charting the dat in cloudwatch it’s clear that the *vast* majority of costs are data from the public internet. That change improved things, but not in a significant way.)

### DIFF
--- a/terraform/loaders.tf
+++ b/terraform/loaders.tf
@@ -87,9 +87,12 @@ module "source_loader_schedule" {
   source   = "./modules/schedule"
   for_each = local.loaders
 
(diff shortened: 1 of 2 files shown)


Mr0grog:
This might blow up; will keep a watch on what actually happens here when the loaders run.

PR #1326, opened 2023-03-02
Expire old build images in ECR

Mr0grog:
We keep built container images in ECR, but have no lifecycle policy and old images keep accumulating! This sets up a policy to keep only the latest 10 images. I think this is probably plenty — we rarely roll back, and have never needed to do so by more than one build.

(Alternatively, we can expire by date pushed, but I think that means if we ever fully pause development, that could actually expire *every* image, which is not good. Docs: https://docs.aws.amazon.com/AmazonECR/latest/userguide/LifecyclePolicies.html)

### DIFF
--- a/terraform/ecr.tf
+++ b/terraform/ecr.tf
@@ -4,6 +4,27 @@
 # store the container images. Images are built and published to these
 # repositories by GitHub Actions.
 
+locals {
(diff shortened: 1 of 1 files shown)


astonm (approved):
👍🏽  Nice. 10 sounds like a good amount.

Issue #1327, opened 2023-03-02
Try removing NAT Gateways in AWS


Mr0grog:
Earlier this week I discovered over 50% of our AWS bill was coming from NAT Gateway fees, and eliminated *most* of that by moving the loaders from our private network into our public network, where they can move data over an Internet Gateway instead (free, vs. billed-per-byte on NAT Gateways). The vast majority of our traffic was raw data we are pulling in from external services. (See: https://github.com/usdigitalresponse/univaf/compare/7b95edb...dad1843)

However! The NAT Gateways are currently still used when data is sent to the API servers (which are in the private network) and when those servers send data out to external services (e.g. to Sentry, Datadog, etc.) or to S3 (when archiving historical data every night). We’re not sending a lot of data over them at this point, but each NAT gateway costs ~$30/month, and we have two (one for each private network). It’s probably safe to move the API servers and the daily data snapshot job to the public network (we need to make sure the security group they are using makes sense). Then we can just eliminate the NAT Gateways entirely (the only thing in the private networks will be the database, and it doesn't need to talk to things outside our VPC, so it won’t need a gateway.

I plan to attempt the above changes tonight (if I’ve got something wrong here, the service will go down until I reverse it, and that would be bad during the day).

**Some background if you are an AWS networking rube like me:**

We have one **VPC**, which sort of a group of networks that all communicate together.

We have 4 **subnets**, which are actual network address spaces where you can run services.
- 2 “public” subnets — each in a different availability zone. When we say “public” here, we mean it has an Internet Gateway (allowing services on it to send/receive data to the internet) and that services on this subnet are automatically given public (visible to the global internet) IP addresses when they are launched.
    - We used to run the bastion in one of these, but otherwise nothing automatic actually ran in the public subnets.
    - Now the loader tasks run in the public subnets.
- 2 “private” subnets — each in a different availability zone (matching the public subnets). When we say “private” here, we mean it has a NAT Gateway (instead of an Internet Gateway) and that services on this subnet are **not** automatically given public IP addresses (but a service on it can still be explicitly given a public IP, which it turns out we have always done for all ECS tasks!
    - The API server runs here.
    - The daily data snapshot task runs here.
    - The loader tasks *used* to run here.
    - The RDS database runs here.

By default, traffic can’t move into/out of/between subnets. But our VPC has a basic routing rule that lets traffic move between any subnets within our VPC. (So any service in any of our subnets can talk to any service in any other subnet.)

[...]

PR #1329, opened 2023-03-03
Ensure new API server can talk to DB

Mr0grog:
Step 1.1 of #1327.

In #1328, I changed the security group the API server tasks use, but forgot to give that security group access to the database! This should fix the issue.

### DIFF
--- a/terraform/db.tf
+++ b/terraform/db.tf
@@ -21,6 +21,7 @@ module "db" {
   subnet_ids = aws_subnet.private[*].id
   ingress_allow_security_groups = compact([
     aws_security_group.ecs_tasks.id,
+    aws_security_group.api_server_tasks.id,
     aws_security_group.bastion_security_group.id
   ])
 }


Mr0grog:
**Lesson learned:** if a security group is needed to be able to access something, you need to create it and add it as something with access and apply those changes, *then wait for those*, then apply a second change that makes services actually *use* the security group. Terraform will not manage the timing between those things correctly on its own.

(This required a second, manual apply in Terraform to get the API servers to run without crashing because they couldn't talk to the DB. OTOH, ECS’s built-in blue-green deploy meant there wasn’t any downtime as a result; it just never rolled forward to the new API servers. 🎉)

This required a second a

Mr0grog:
And looking great in CloudWatch. NAT Gateway traffic has gone down to 0, but everything is still working:

Screen Shot 2023-03-02 at 10 06 40 PM

PR #1330, opened 2023-03-03
Remove NAT Gateways from private subnets

Mr0grog:
Step 2 of #1327. This removes the NAT Gateways that have been providing internet access to the private subnets, since we no longer run any services that need internet access in those subnets. They are now only for running services that are completely scoped to our VPC.


Mr0grog:
🤞

PR #1331, opened 2023-03-03
Stop using deprecated security groups

Mr0grog:
Step 3 of #1327. Now that we've rejiggered our security groups, we need to stop using them. After this applies, then we can actually delete them.


Mr0grog:
RDS is now updated and still working fine. ✅

PR #1346, opened 2023-03-07
Randomize test order

### LINKED ISSUE #1345: Use new `randomize` option in Jest
In v29.5.0, Jest added an option to randomize the order of tests. Doing so is a good practice to make sure you stay honest with keeping tests independent and less error-prone, but it breaks some of our tests right now!

We should be using it, but there appear to be errors in:
- Loader:
    - `walgreens.test.js` → [“Walgreens SMART Scheduling Links API › should set availability to NO if no slots are free”](https://github.com/usdigitalresponse/univaf/blob/9dd276c576e494aba14b3ab0ba1cffab7f8a7587/loader/test/walgreens.test.js#L76-L96)
- Server:
    - `db.test.ts` → [“db.addExternalIds › updating locations › does not allow a colon in the externalId”](https://github.com/usdigitalresponse/univaf/blob/e56138447e69673eb403caa59002629755197190/server/test/db.test.ts#L699-L709)
- (Common seems ok)

Mr0grog:
Jest recently added an option to randomize the order of tests. In general, that's a good practice that can help ensure we make tests more independent and robust, so we should be using it.

Fixes #1345.

### DIFF
--- a/common/jest.config.js
+++ b/common/jest.config.js
@@ -11,6 +11,7 @@ module.exports = {
   },
   testMatch: ["**/test/**/*.test.(ts|js)", "!**/dist/test/**/*"],
   testEnvironment: "node",
+  randomize: true,
   reporters: ["default", "github-actions"],
   coverageReporters: ["text", "html"],
 };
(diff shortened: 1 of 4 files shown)


Mr0grog:
I was previously seeing an error in `db.test.ts`, but am having trouble reproducing it now. 🤷

PR #1351, opened 2023-03-10
Remove no longer needed DB ingress security

Mr0grog:
In #1349, I flipped how security groups work with RDS, and services that need access to the DB now include a security group that the DB outputs. We no longer need to specify allowed services as inputs.


Mr0grog:
Here we go 🤞

PR #1366, opened 2023-03-14
Remove Render configuration & AWS resources

Mr0grog:
Just some lingering cleanup from our previous experience on Render. This removes the blueprint file that is no longer used, and some of the AWS resources that were specific to Render.


Mr0grog:
This also does not actually remove the Render-related domains and CloudFront because we already did that through configuration. The code supporting those that is being removed here is already effectively dead code.

PR #1370, opened 2023-03-17
Repair the Rite Aid scraper

Mr0grog:
Our Rite Aid scraper broke a few days ago, and it seems like Rite Aid made some small but complicated changes to their web scheduler’s API.

The main thing is that you can no longer request more than 100 locations at a time. There is no pagination mechanism, so I needed to re-think the zip codes and radii we query with in order to make sure we actually get all the results. Even worse, I realized partway through that the API is not actually returning all the locations in the given radius (!). I spent way too much time trying to figure out what’s going on, and I don’t get it. I think the actual effective radius you get might have something to do with the density of stores in a given area, but I really can’t understand what it’s doing from the outside. That meant a lot of time playing with different configurations of zipcodes and radii.

As a result, I built a working zipcode list for New Jersey that seems to return all the stores, but for the other states, I checked the number of stores it was getting against the [store list on riteaid.com](https://www.riteaid.com/locations/index.html) and left it alone if it got everything or added a `FIXME` comment and removed it from the `--states` option in the cron job configuration if not. I’ll have to come back to those later.

There were some other minor changes:
- Rite Aid no longer has any stores in Colorado (this has been true for more than a year, but querying in an empty state only started causing problems now).
- The format of some fields changed.
- Added bivalent booster support.

Fixes https://usdr.sentry.io/issues/2420647478

---

**Side note:** I’m going to merge this to get things working again, but since slot-level information is no longer super critical, maybe it’s time to turn off this scraper and just use the other, less detailed Rite Aid APIs.


loader/test/riteaid.scraper.test.js
```diff
@@ -145,7 +145,7 @@ describe("Rite Aid Scraper", () => {
         },
       });
 
-    const result = await checkAvailability(() => {}, { states: ["NJ"] });
+    const result = await checkAvailability(() => {}, { states: ["CT"] });
```
Mr0grog, line 148: I changed all these tests to "CT" since NJ now requests multiple zip codes. CT still only uses one, and that keeps things straightforward in the tests.

terraform/cron-loaders.tf
```diff
@@ -36,10 +36,14 @@ locals {
...
-    riteAidScraper = { schedule = "cron(0/10 * * * ? *)" }
+    riteAidScraper = {
+      schedule = "cron(0/10 * * * ? *)"
+      # FIXME: a bunch of states are broken in the scraper right now. Remove
+      # this narrow list when the unlisted states are fixed.
+      options = ["--states", "CT,DE,MA,NH,NJ,NV,OR,VT"]
+    }
     riteAidApi = {
       schedule = "cron(0/30 * * * ? *)"
-      options  = ["--states", "CA,CT,DE,ID,MA,MD,MI,NH,NJ,NV,NY,OH,OR,PA,VA,VT,WA"]
```
Mr0grog, line 42: This allow list of states was only used because the public API stopped supporting Colorado a while back, before it was clear those stores were permanently closed. Since there are definitely closed now, I removed them from the list of default states for Rite Aid, and this list is no longer needed.

(But now we need a list for the scraper, which uses the private API and is broken for some states. 🙃)

PR #1395, opened 2023-03-28
Add SMART SL source for Rite Aid

### LINKED ISSUE #159: Implement SMART loader for Rite Aid
We currently load data via Rite Aid’s API at https://api.riteaid.com/digital/Covid19-Vaccine/ProviderDetails. They’re releasing a SMART SL API soon (scheduled for 6/1), and we should implement against it once it’s available.

Update: new SMART SL API is at: https://api.riteaid.com/digital/vaccine-provider/$bulk-publish
You can also find their staging server at: https://apicar.riteaid.com/
You can also find their QA server at: https://apisqa.riteaid.com/

Mr0grog:
This adds a SMART Scheduling Links API source (`riteAidSmart`) for Rite Aid, with the ultimate goal of replacing/deprecating the “API” source (`riteAidApi`). The two have *roughly* the same granularity of info, but the “API” source requires authentication and the SMART SL one does not.

I’ve put off implementing this for a while because the existing source already worked fine and Rite Aid’s SMART SL API was a little buggy (it still is, but the worst parts have been fixed and the remaining bugs aren’t especially problematic for our needs). Recently, the existing “API” source has been encountering HTTP 403 (access forbidden) errors every so often, which is not really great. It’s unclear whether this is a problem on their side, a ramp-up to turning off the API, or something else (maybe one of the AWS instances Fargate allocates for some runs has been blocked?). In any case, it seemed worth trying out the SMART SL API instead to see if things work better there.

This adds a few pieces of data that were missing from the API source (geographic positions, VTrckS PINs) and is missing some data the API source has (no update times, although it appears to potentially be live; no count of *busy* slots, just free ones). These are tradeoffs, but I think they are probably fine. See https://github.com/usdigitalresponse/univaf/issues/159#issuecomment-1476723402 for details.

Fixes #159.


loader/test/support/schemas.js
```diff
@@ -44,6 +44,7 @@ const capacitySchema = {
     unavailable_count: { type: "integer", minimum: 0 },
     products: productsSchema,
     dose: { type: "string" },
+    booking_url: { type: "string", format: "uri" },
```
Mr0grog, line 47: It turns out other sources have been including this, but didn’t have test data that covered it. 🤷

Mr0grog:
The immediate impetus for this (the 403 status codes) turns out to probably still be a problem with this API (see https://github.com/usdigitalresponse/univaf/pull/1397#issuecomment-1497760234), but I still plan to try and clean this up and merge it next week, as long as it’s not too much remaining work. It’s worth having in the end, even if not high priority.

Mr0grog:
MAJOR caveat discovered once I got this in production: it does not list locations with no open slots. So those locations won’t get updated to indicate they don’t have slots. This is a huge problem, and I will probably turn this back off rather than turning off the old API that requires authentication.

PR #1397, opened 2023-03-29
Retry Rite Aid requests on 403 status codes

Mr0grog:
The Rite Aid API and scraper sources both occasionally encounter a raft of 403 status codes and fail (anywhere from 1/hour to 1/day). Under the assumption that we’re hitting a rate-limiting/abuse measure, this change sets us up to retry those requests rather than failing immediately, but with a greater than usual delay.

This is an experiment! It might not fix the issue — an equally likely possibility is that because we run on Fargate, each loader run comes from one of a wide set of IP addresses, and some of them have been blocked by Rite Aid. If that's the case, retrying won't do us any good. (OTOH, it should be obvious that this is the case if the change here has zero impact on the rate at which we see errors, and then we take appropriate action, like logging clearer errors and failing faster when we hit this scenario.)

### DIFF
--- a/loader/src/sources/riteaid/api.js
+++ b/loader/src/sources/riteaid/api.js
@@ -1,3 +1,4 @@
+const got = require("got");
 const { DateTime } = require("luxon");
 const Sentry = require("@sentry/node");
 const geocoding = require("../../geocoding");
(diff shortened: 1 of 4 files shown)


Mr0grog:
UPDATE on “This is an experiment! It might not fix the issue…”

Screen Shot 2023-04-05 at 9 01 19 AM

This doesn’t appear to have had any impact at all, so the most likely cause is that this isn’t rate limiting; it’s just that our jobs sometimes run from a blocked IP. The better fix here is to just fail faster and stop. (We could also set up a proxy, but a low level of occasional failures is really fine.)

PR #1398, opened 2023-03-31
Fix Rite Aid 403 retry exception

Mr0grog:
I made a mistake in #1397 by assuming every time the `calculateDelay` function runs for retries, the error involved an actual HTTP response. Some are network problems where the client is unable to communicate with the server at all. (I tested 403 responses, but did not test this!)

This also consolidates the logic in one spot, since it was obviously more complicated than I gave it credit for (I didn’t bother before since it was experimental. That was wrong.).

Fixes https://usdr.sentry.io/issues/4048563618

### DIFF
--- a/loader/src/sources/riteaid/api.js
+++ b/loader/src/sources/riteaid/api.js
@@ -121,7 +121,7 @@ async function queryState(state, rateLimit = null) {
       // than normal delay.
       statusCodes: [...got.default.defaults.options.retry.statusCodes, 403],
       calculateDelay({ error, computedValue }) {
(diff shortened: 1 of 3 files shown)


Mr0grog:
I also made a logic tweak here: the previous version had a minimum delay of 10 seconds, and would retry just as many time as for any other sort of response. Now it only retries once on 403 status codes, but has a longer delay of 30 seconds.

I figured having a longer delay would be better to escape rate limiting/abuse issues, and retrying a whole bunch of times doesn’t really make sense in this case (also, that keeps the longer delay from getting out of control!).

Issue #1459, opened 2023-05-02
Add stale data warnings


Mr0grog:
The CVS issue in #1458 and other similar issues of API data freshness in the past have always been discovered by chance, and by us paying regular attention to the data. I was a little slow on catching CVS because the system has moved more towards pure maintenance, and that’s a problem. We need some alerts when data ceases to be fresh.

Some potential approaches:
- Places to check:

    1. **The loader can check all the records it outputs and log issues.** I like that this is close to where things happen, and is very immediate. OTOH, it could be *too* sensitive. Not sure. This also lets us summarize across all the data surfaced by a given source, and at a given time (doing the checks elsewhere might wind up including locations surfaced by that source in the past, but that are no longer surfaced by that source — e.g. this might be a big problem for Prepmod, which has a lot of one-off/two-off *events*).

    2. **The server’s `/api/edge/update` endpoint could check every record it receives.** We already have [a freshness timeframe encoded in the server](https://github.com/usdigitalresponse/univaf/blob/26212136b89380971fdcf35251fbdef7016465da/server/src/db.ts#L28-L33), and this puts these freshness checks in a nearby place. This is also just as immediate as the loader. OTOH, the server doesn’t know about a whole loader run, so this can’t summarize across a source (other than keeping a running average or something).

    3. **A scheduled job (probably 1–4× /day) could calculate statistics across the whole database.** It lets us do lots of summary statistics, but can't tell whether an availability record is stale because the source no longer outputs info for that location. It also needs configuration or coding to know which sources to configure (since some are deprecated and no longer in use). However, it can catch issues caused by the loader never running or never surfacing any data at all for a given source, which the other approaches would miss.

- What constitutes “stale?”
    - `<record>.availability.valid_at` is vastly older than `<record>.availability.checked_at` or the current time.
    - `<record>.availability.slots` or `<record>.availability.capacity` are present, but all the entries are in the past.
    - ⚠️ Question: do we have at least one of these for every source currently in use?

- How do we track it?
    - Log a warning to Sentry. This is pretty binary, and adjusting the threshold means code changes and re-deployment.
    - Track freshness (days/hours/seconds out of date) as a metric in Datadog. Then we can just adjust the threshold for alarms without deploying anything. We can also track different different summary stats by source (average, max, min).

Mr0grog:
Of existing, actively used sources, can we get freshness info (and how)?

[...]

Issue #1458, opened 2023-05-02
Determine course of action for CVS (API stopped updating)


Mr0grog:
The CVS SMART SL API is still operating, but it has stopped returning new, up-to-date data. I've messaged their lead dev on the FHIR discussion list, and will wait a bit to see if I hear back. If not, it may be time to revive the scraper (I took a quick look and a few things need to change — result distance appears to always be 35 miles, regardless of what you request; vaccines are specified by NDC code), but it should be doable.

Like we agreed for future Rite Aid scraping, we probably want to limit this to New Jersey and Alaska, where we have active government partners. (Scraping is high impact for both us and the scraped server, plus lots of effort to maintain. Things are less of an emergency these days, so we want to limit the scope of any scraping where possible.)

Mr0grog:
Initial feedback from partners is that this is not mission-critical — what’s critical today is information about the locations (which we still have) since most places always have open slots in practice now.

Dropping the priority on this.

Mr0grog:
I’ve successfully gotten in touch with vaccines.gov and BCH folks, who are following up with CVS and Kroger (I was never able to get in touch with CVS or Kroger developers who I worked with before, but the people running vaccines.gov ought to be able to grab the pharmacies’ attention).

Since we are currently planning to shut down UNIVAF in a month, I don’t think it makes sense to take this any farther.

PR #1465, opened 2023-05-05
Add scraper for Albertsons

Mr0grog:
Albertsons recently switched off MHealth and onto their in-house booking system for COVID vaccines. This implements a new scraper for the in-house booking website, although I've only calculated queryable zip codes for New Jersey and Alaska.

I *probably* won't be working hard to implement a scraper for CVS or Kroger (see #1458, #1459) since those two have official APIs in use by the federal government and might possibly get fixed. Albertsons does not, however, so I went ahead and wrote a quick scraper for it. That said, our partners have indicated that the freshness issue isn’t a horrible deal-breaker at this point, so it’s not mission-critical.

This hits the JSON API that powers the booking site, but there are a few important notes:

- The API doesn't indicate whether appointments are available, but appears to only return locations where appointments are available. To handle that, we mark all locations returned by the API as available, then look at our database of all Albertsons locations for any that weren't returned and mark those as unavailable.

- The API doesn't let you do an OR query for multiple vaccines (only an AND query). That means we'd have to make a ton of requests to figure out which products are available, so we don't (fortunately we can fall back on the CDC API for that data, like we do for CVS and some other stores).

- The set of zip codes to query are hand-tuned to fit a 50 mile radius. It doesn't *seem* like there's a limit on the number of results, and in practice we seem to be getting data about every location in the states I set things up for. But it won't be obvious if we suddenly start getting a subset of locations.


loader/src/sources/albertsons/index.js
```diff
@@ -692,7 +530,7 @@ function formatLocation(data, validAt, checkedAt) {
   return {
     name,
     external_ids,
-    provider: "albertsons",
+    provider: storeBrand.provider_id,
```
Mr0grog, line 533: I think this is what this was always supposed to be, so I went ahead and "fixed" it. (Even though this scraper is deprecated now.)

terraform/cron-loaders.tf
```diff
@@ -35,7 +35,7 @@ locals {
     # FIXME: Albertsons needs a new scraper. The system the current one was
     # written for has been abandoned by Albertsons, so all info in it is stale.
     # See https://github.com/usdigitalresponse/univaf/issues/1460
-    # albertsons     = { schedule = "cron(6/10 * * * ? *)" }
+    # albertsonsScraper     = { schedule = "cron(6/15 * * * ? *)" }
```
Mr0grog, line 38: Will uncomment this in a follow-on PR once the new code is deployed.

PR #1468, opened 2023-05-08
Remove specialized Albertsons booking URLs

Mr0grog:
In mHealth Coach, some locations would have different booking URLs for different products (or sets of products). The way those worked in mHealth Coach kept evolving, so we added an ad-hoc system for supporting multiple booking URLs as properties of a location's `meta` field (why didn't we do this as part of the *availability* object's `meta` field? I do not remember. It seems like it would have made more sense.). You can see how that was done here: https://github.com/usdigitalresponse/univaf/blob/a73e3912f36cd5a92b8069695e6149acb784d667/loader/src/sources/albertsons/mhealth.js#L163-L172

and here: https://github.com/usdigitalresponse/univaf/blob/a73e3912f36cd5a92b8069695e6149acb784d667/loader/src/sources/albertsons/mhealth.js#L549-L566

Since Albertsons no longer uses mHealth Coach and these are part of the location rather than the availability, we need to manually clear them out with a migration.

We also need the new scraper to update location info, since it has a new top level `booking_url` property and possibly other new values as well (like store hours).

### DIFF
--- a/loader/src/sources/albertsons/scraper.js
+++ b/loader/src/sources/albertsons/scraper.js
@@ -372,7 +372,7 @@ async function checkAvailability(
         });
         location = formatLocation(apiLocation, checkedAt);
       });
-      handler(location);
+      handler(location, { update_location: true });
       results.push(location);
       storeNumbers.add(location.meta.albertsons_store_number);
     }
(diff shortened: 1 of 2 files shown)


Mr0grog:
This updated 1,893 locations.

PR #1508, opened 2023-05-24
Make sure all loader sources are directories

Mr0grog:
This is just a little cleanup refactor so that all sources are structured the same. I also have some bits of uncommitted code that's been floating around on my local drive for a while that goes along with some sources that previously weren't in directories, and since I'd like to commit it for posterity, having a directory to put it in is nice.


Mr0grog:
Whoops, lost track of this yesterday and forgot to merge!

PR #1510, opened 2023-05-25
Add store list scrapers for CVS, Kroger

Mr0grog:
This is pretty analogous to the scrapers we wrote a while back to get all Albertsons stores. I wrote these during the mess around #1458 and #1459, and have had them just sitting uncommitted for a while. In the end, they weren't worth working more on since we are ramping things down, but they would probably be useful for anyone who might want to make use of the scrapers and tools in this codebase, so I felt it was worth committing for posterity and archival purposes.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -9,6 +9,7 @@ terraform
 
 **/fixtures/**/*.json
 **/__fixtures__/**/*.json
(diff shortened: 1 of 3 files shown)


Mr0grog:
May as well merge this in.

PR #1511, opened 2023-05-26
Make server scripts TypeScript, remove duplicated code

### LINKED ISSUE #199: Clean up merge scripts
We have some tooling for merging and removing duplicates in the `server/scripts/` directory (see also #157 and #196). Some of the code is pretty hairy, it’s written in JS instead of TS, and there are no tests. All the merging logic also lives directly in the scripts rather than in the main codebase (e.g. in `db.ts`), which is probably not where it should belong.

It would be good to clean these up by:

- [ ] Making them TypeScript. (We probably need to use the `bin` field in `package.json` or have a `script` in `package.json` that compiles & runs them on the fly to make using them straightforward.)
- [ ] Adding tests! 😬 
- [ ] Moving relevant code into the main server codebase. Since these both do a kind of funky thing by creating a “plan” that can be printed and optionally executed, I’m not sure how much is easily transferable.

This is not super-high priority, though. The most valuable thing would probably be tests. :)

Mr0grog:
This does a little not-entirely-necessary cleanup by finally making the server scripts TypeScript, which also lets us move a bunch of duplicated code to the `univaf-common` module more easily. We can also now use the same tools for databases in the scripts as in the main server.

Fixes #199.

### DIFF
--- a/common/src/address.ts
+++ b/common/src/address.ts
@@ -171,3 +171,25 @@ export function parseUsAddress(address: string): UsAddress {
     zip,
   };
 }
+
+/**
+ * Format a US ZIP code into its canonical 5-digit form (plus an optional
+ * 4-digit suffix). This is useful since a lot of systems return ZIP codes
+ * without leading zeroes.
+ */
+export function formatZipCode(zipCode: string | number): string {
+  if (typeof zipCode === "number" && !Number.isInteger(zipCode)) {
+    throw new SyntaxError(`ZIP code is not an integer: "${zipCode}"`);
+  }
+
+  const parts = zipCode.toString().split("-");
+  if (parts.length > 2) {
+    throw new SyntaxError(`ZIP code has multiple dashes: "${zipCode}"`);
+  }
+
+  let result = parts[0].padStart(5, "0");
+  if (parts.length > 1) {
+    result = `${result}-${parts[1].padStart(4, "0")}`;
+  }
+  return result;
+}
--- a/common/src/address.ts
+++ b/common/src/address.ts
@@ -1,0 +1,173 @@
+import { ParseError } from "./exceptions";
+
+const MULTIPLE_SPACE_PATTERN = /[\n\s]+/g;
+const PUNCTUATION_PATTERN = /[.,;\-–—'"“”‘’`!()/\\]+/g;
+const POSSESSIVE_PATTERN = /['’]s /g;
+
+const ADDRESS_LINE_DELIMITER_PATTERN = /,|\n|\s-\s/g;
+const ADDRESS_PATTERN =
+  /^(.*),\s+([^,]+),\s+([A-Z]{2}),?\s+(\d+(-\d{4})?)\s*$/i;
+
+// Common abbreviations in addresses and their expanded, full English form.
+// These are used to match similar addresses. For example:
+//   "600 Ocean Hwy" and "600 Ocean Highway"
+// They're always used in lower-case text where punctuation has been removed.
+// In some cases, the replacements *remove* the abbreviation entirely to enable
+// better loose matching (usually for road types, like "road" vs. "street").
+const ADDRESS_EXPANSIONS: [RegExp, string][] = [
+  [/ i /g, " interstate "],
+  [/ i-(\d+) /g, " interstate $1 "],
+  [/ expy /g, " expressway "],
+  [/ fwy /g, " freeway "],
+  [/ hwy /g, " highway "],
+  [/ (u s|us) /g, " "], // Frequently in "U.S. Highway / US Highway"
+  [/ (s r|sr|st rt|state route|state road) /g, " route "],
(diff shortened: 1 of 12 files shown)


astonm:
Yay!
image

common/src/address.ts
```diff
@@ -0,0 +1,173 @@
...
+    match[1].replace(PUNCTUATION_PATTERN, "") === "" ||
+    match[2].replace(PUNCTUATION_PATTERN, "") === "" ||
+    match[4].replace(PUNCTUATION_PATTERN, "") === "";
+  if (invalidMatch) {
+    throw new ParseError(`Could not parse address: "${address}"`);
+  }
+
+  const zip = match[4];
+  if (zip.split("-")[0].length < 5) {
+    throw new ParseError(`Invalid ZIP code in address: "${address}"`);
```
Mr0grog, line 164: Minor note: this used to log a warning and set `zip` to `undefined` instead of throwing. However, the common module doesn’t have access to the same Sentry logging tools as the loader, so there wasn’t an easy way to do that (arguably there should be some common logging infrastructure in the univaf-common module, but that’s a whole other ball of wax).

Gracefully handling invalid zip codes was mainly an issue with The old MHealth Coach system that Albertsons used, and which they don’t use anymore, so this shouldn’t pose an actual issue in practice.

server/scripts/availability_dump.ts
```diff
@@ -112,20 +112,24 @@ async function getAllBucketObjects(options) {
  * performs *terribly* and is often worse than no chunking at all.
  */
 class BufferedStream extends stream.Transform {
+  #buffer: Buffer = null;
+  #size = 0;
+  #offset = 0;
```
Mr0grog, line 117: Fancy-pants private properties! Don’t know why I wasn’t using these before; we don’t need TypeScript for them in the current version of Node.js.

server/scripts/merge-locations.ts
```diff
@@ -4,150 +4,59 @@
...
-  // If there are multiple lines and it looks like the first line is the name
-  // of a place (rather than the street), drop the first line.
-  if (lines.length > 1 && !/\d/.test(lines[0])) {
-    lines = lines.slice(1);
-  }
-
-  if (line != null) {
-    lines = lines.slice(line, line + 1);
-  }
+export const db = createDbClient("Scripts");
```
Mr0grog, line 51: 🎉 

---

Side note: it’s still nice to use a separate client instance here rather than the global one from `src/db.ts` because the `main()` function will close the connection. The lifecycle expectations for the two instances are a little different, so keeping them separate makes maintaining tests easier.

server/scripts/remove-duplicate-locations.ts
```diff
@@ -106,22 +112,27 @@ function mergeGroupsWithCommonLocations(groups) {
...
+function planChanges(mergeGroups: Set<ScriptLocation>[]) {
   // Plan merges for each group.
   const plans = [];
   for (const group of mergeGroups) {
     // Sort by oldest first, so that the oldest becomes the target all the
     // others merge into.
-    const sorted = [...group].sort((a, b) => a.created_at - b.created_at);
+    const sorted = [...group].sort(
+      (a, b) => a.created_at.getTime() - b.created_at.getTime()
+    );
```
Mr0grog, line 135: Adding `.getTime()` is not technically required here, but TypeScript doesn’t like subtracting `Date` instances. This is, in some very technical sense, more correct. 🤷

server/scripts/remove-duplicate-locations.ts
```diff
@@ -161,22 +167,36 @@ async function doChanges(plans, persist = false) {
...
+  ..._
+}: {
+  unpad?: boolean;
+  commit?: boolean;
+  system?: (string | number)[];
+}): Promise<void> {
+  assert.ok(
+    system.every((s) => typeof s === "string"),
+    "Systems must all be strings"
+  );
```
Mr0grog, line 183: This has never been an issue in practice, but the TS compiler reminded me that Yargs arrays are always `string|number`; you can’t really stop it from parsing the values of an array option as numbers, AFAICT.

server/scripts/remove-duplicate-locations.ts
```diff
@@ -239,7 +251,7 @@ if (require.main === module) {
               and CVS IDs, use: "--system kroger --system cvs".
             `.trim(),
           }),
-      handler: main,
+      handler: (args) => main(args),
```
Mr0grog, line 254: This is a little weird, but makes testing `main()` easier. Yargs’s types really want the handler to take an instance of `yargs.ArgumentsCamelCase<your_options>` or `yargs.Argv<your_options>` (and not just some compatible subset), but creating one of those when calling `main()` in a test kinda sucks. This is a convenient way to split the difference. TypeScript still makes sure things are compatible, but doesn’t make strict requirements about the exact function arguments.

PR #1523, opened 2023-05-30
Slim down the production service

Mr0grog:
As we move towards shutdown, this simplifies the service to using only one availability zone and reduces the loader schedules so that none run any more often than every 15 minutes.

### DIFF
--- a/terraform/cron-loaders.tf
+++ b/terraform/cron-loaders.tf
@@ -30,10 +30,10 @@ locals {
     },
     waDoh             = { schedule = "cron(3/30 * * * ? *)" }
     cvsSmart          = { schedule = "cron(3/15 * * * ? *)" }
-    walgreensSmart    = { schedule = "cron(2/10 * * * ? *)" }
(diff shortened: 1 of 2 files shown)


Mr0grog:
We should probably expect some system errors when load balancers disappear in the middle of loader runs here.

Mr0grog:
This didn’t actually work:
- Changing the AZs forced all our subnets to be recreated, but due to the way subnets are set up, the system tried to use the same subnet IP blocks, and you can’t have multiple subnets with the same IP block.
- The database uses a subnet *group*, which can’t be configured with < 2 subnets (it uses one subnet from each AZ). https://github.com/usdigitalresponse/univaf/blob/4331ca9d65487be82eab8892f54cb1f1e18a5407/terraform/modules/rds/main.tf#L44-L48

We could do something complex like adding a variable to only *use* N of our total M availability zones, but that just feels like adding complexity, and the purpose of this was to reduce complexity. So I’m going to drop the attempt to use fewer AZs, and just do the slower loader schedule in #1524.

PR #1544, opened 2023-06-06
HOTFIX: Support nested arrays in Walgreens

Mr0grog:
Walgreens's SMART SL manifest file currently lists the slots as a nested array of output items inside the main `output` array, which is not correct, and has been causing us to list all Walgreens as unavailable.

For example, we are seeing a manifest like:

```js
{
  "transactionTime": "2023-06-06T03:21:31.360Z",
  "request": "...",
  "output": [
    { "type": "Location", "url": "..." },
    { "type": "Schedule", "url": "..." },
    // These items should not be in a sub-array; they should just be in the `output` array directly.
    [
      { "type": "Slot", "url": "...", "extension": { "state": ["AK"] } },
      { "type": "Slot", "url": "...", "extension": { "state": ["AL"] } },
      // etc...
    ]
  ]
}

```

This hacks in a quick fix: we just flatten the `output` array before handling it. I've also alerted the maintainer of the Walgreens API to the issue, so hopefully the underlying system gets fixed soon.

Discovered while looking into to some transient errors from Walgreens after merging #1531. Not sure whether it’s new with the authorization fix this morning on Walgreens's side, or if it's been happening for a while. :\

### DIFF
--- a/loader/src/smart-scheduling-links.js
+++ b/loader/src/smart-scheduling-links.js
@@ -120,7 +120,10 @@ class SmartSchedulingLinksApi {
         ...this.httpOptions,
         url: this.url,
       }).json();
+      // FIXME: We should validate the schema of the response.
       this.manifest = { time: Date.now(), data };
+      // HACK: This is a [temporary?] workaround for bad data from Walgreens.
+      this.manifest.data.output = this.manifest.data.output.flat();
     }
     return this.manifest.data;
   }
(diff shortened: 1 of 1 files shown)


loader/src/smart-scheduling-links.js
```diff
@@ -120,7 +120,10 @@ class SmartSchedulingLinksApi {
         ...this.httpOptions,
         url: this.url,
       }).json();
+      // FIXME: We should validate the schema of the response.
```
Mr0grog, line 123: I’m going to add some `assertSchema()` calls tomorrow or later this week, since this highlights the fact that we should have made sure we were alerted that the data was malformed instead of processing it incorrectly.

PR #1563, opened 2023-06-13
Absurdly unnecessary cleanup refactor: Make loader sources async iterators

### LINKED ISSUE #55: Loaders should take the callback as the last argument
We should match common JS conventions, and `checkAvailability(handler, options)` definitely has the handler/callback in the wrong place.

Mr0grog:
This fixes #55 by getting rid of the callback argument for `checkAvailability()` entirely. One of the weird and messy things in the loader sources is that their main function (`checkAvailability()`) takes a callback to run for every found location (so we get results outputting progressively and reducing load to the server receiving them) but also returns an array of all the found locations (mostly for easy testing). IIRC, I was originally going to make it return a stream instead, but the mechanics of that turned out to be pretty complicated. In #55 I wanted to at least clean it up a bit and position the callback in a more conventional spot, but I actually think async iteration makes the most sense here; it gets us the stream-like stuff with simpler syntax.

Not everything is as clean as I was hoping here; I had been thinking test code would be really simple because we could polyfill the new `Array.fromAsyc()` or `AsyncIterator.prototoype.toArray()` functions, but forgot that the callbacks optionally take a second argument for options about how the location is sent to the database, and so an array of a sources outputs is not necessarily an array of the location objects. So the tests have a new helper to get just the locations.

The CLI module has slightly more code to manage this, too, but I think the overall working of it is slightly simpler, since we can just compose a series of iterators using `stream.compose()` for all the result processing steps.

### DIFF
--- a/loader/src/cli.js
+++ b/loader/src/cli.js
@@ -1,5 +1,7 @@
 "use strict";
 
+const { pipeline } = require("node:stream/promises");
+const { compose } = require("node:stream");
 const Sentry = require("@sentry/node");
 const yargs = require("yargs");
 const { states: allStates } = require("univaf-common");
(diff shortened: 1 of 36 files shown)


astonm (approved):
This refactor looks pretty solid overall, and I think communicates the right ideas about what's happening via the more modern syntax.

Just curious, did you use any tooling to automate the refactor?

Mr0grog:
> Just curious, did you use any tooling to automate the refactor?

I probably should have! But this was pretty much just multi-file find-and-replace with a few regexes across the `loader` directory. In retrospect, Codemods or something probably would have been good to do.

PR #1569, opened 2023-06-15
Add tombstone page

Mr0grog:
This adds a really simple, static “tombstone” site that will replace the API server at https://getmyvax.org this evening when we shut things down. It’s intended to deploy on GitHub pages (replacing the current “demo UI”); we’ll just update the getmyvax.org DNS to point to it later on.

Originally I was going to have separate index and 404 pages, but realized that wasn’t useful or necessary — all we really need is a universal 404 that comes up regardless whether you go to https://getmyvax.org/ or https://getmyvax.org/whatever. I also poked around a bit at using a build script or static site generator to convert a Markdown file to HTML, but with just a single, short, not-too-complex page, that seemed kind of pointless and it doesn’t make it meaningfully easier to tweak or update if we need to do so. So I just wrote some quick and simple HTML by hand.

The layout is really simple:

Screenshot of tombstone page


.github/workflows/deploy-tombstone.yml
```diff
@@ -0,0 +1,43 @@
...
+  deploy:
+    if: github.ref == 'refs/heads/main'
+    runs-on: ubuntu-latest
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v2
```
Mr0grog, line 43: Also switching to the new method of deploying to GH pages, where you build and upload a zip of the site, rather than committing to a special branch.

Issue #1550, opened 2023-06-09
☂️ Shut down production service


Mr0grog:
On June 15, 2023, well be shutting down the production service, but will be keeping the source code available here, as well as the archived data at [archives.getmyvax.org](https://archives.getmyvax.org) and a placeholder page at [getmyvax.org](https://getmyvax.org).

What needs to be done to wrap things up?

- [x] After Monday’s dependency updates, disable Dependabot (maybe leave security updates on?). (#1564)
- [x] Add maintenance status notice to README (#1588)
- [x] Create a home page for https://archives.getmyvax.org with some minimal documentation (how to construct download paths and information about file formats). I’ve been meaning to do this for a while, but it’s more critical now that the exiting docs page will be gone. (#1570)
- [x] Create a tombstone and 404 page for https://getmyvax.org. (#1569)
    - Publish to GitHub pages (instead of the existing demo UI, so this will require changing the [`ui-deploy.yml` actions workflow](https://github.com/usdigitalresponse/univaf/blob/fdcfe041238c39d6ad33898ab4642c1edd037bcd/.github/workflows/ui-deploy.yml).
    - This can probably just replace the [demo UI in the `ui` folder](https://github.com/usdigitalresponse/univaf/tree/main/ui); I don’t think there’s a real reason to keep that code around and it hasn’t been well maintained anyway.
- [x] Do a final pass on docs; make sure it’s *reasonably* clear how to fork and deploy yourself.
- [x] Clean up dangling PRs and issues.
    - We should probably close all lingering PRs.
    - I plan to leave issues that someone forking and re-deploying might want to do (e.g. #202) but close others that just aren’t relevant anymore (I’ve already done a bunch of these).
- [x] Make a copy of the existing Terraform files in `/terraform/deprecated` (similar to what we did during the failed switch to Render; see #930) and add a README there that describes the fact that those files represent a production deployment, while the top-level `/terraform` directory is the archived/shut-down deployment. (#1565)
- [x] Update loader schedules so they don’t run at or after `2023-06-16T00:00:00Z`. (e.g. a schedule like `cron(1/15 * * * ? *)` should become `cron(1/15 * 1-15 6 ? *)`). (#1566)
- [x] Wait until the start of June 16 UTC.
- [x] Switch DNS for `getmyvax.org` to point to tombstone page on GitHub pages. (#1571)
[...]

PR #1570, opened 2023-06-15
Add docs site for archives.getmyvax.org

Mr0grog:
We’ve needed to have some docs around https://archives.getmyvax.org for a while and never gotten around to it. Now that the API and its accompanying docs are disappearing, I wanted to add some docs that live directly on `archives.getmyvax.org` (so whatever else happens to other sites, the docs stay alongside the archive data).

The docs page got quite long once I’d pasted in all the schemas and sources and ID systems and so on, so I split it up into a few pages and used MkDocs to generate a site from the markdown files.

All the source for the docs site is in the `archives` directory. There’s a workflow to build the site and upload it to the `/docs` folder in the `univaf-data-snapshots` S3 bucket where we serve the archives site from.

This got a bit in the weeds on a few counts:

- Tooling: I poked around at a few docs generators that I had and hadn’t used before. Ultimately I felt like MkDocs (and the Material theme for it) gets us something pretty nice without a giant mess of configuration and without *too* much extra files and stuff (no React, etc. etc. etc. [TIL Docusaurus 2 is React-based!]). It’s Python, though, which makes the set up weird vs. the rest of the stuff in this repo.

    I also played around with 11ty for a very minimal, JS-based site, but quickly found myself in the weeds on layout concerns, anchors for headings, and so on. You don’t really get much out of the box with it. We can build something more minimal, but I’m not sure that’s really important.

- S3 Website Serving: I decided not to turn it on, which means URLs are not as nice, and there’s a weird index redirect thing I’m doing.

    I’ve never set up S3 in website mode before, and my lesson from today is that it’s not that complicated, but then if you want to force people to go through CloudFront (which we do) and still secure the bucket it’s a *little* complicated, and then because we are using a Cloudposse Terraform module to manage CloudFront, but *not* to create and manage the bucket itself, it’s really complicated. Did not feel worth the time to figure out how to do it right, or to unwind our usage of Cloudposse.

    The downside is that this means browsing to `/docs/` does *not* get you `/docs/index.html` (this still works for the root `index.html` file, just not for subdirectories). So I had to flip MkDocs's settings to generate `<page>.html` instead of `<page>/index.html`.

    Since Mkdocs generates a bunch of files, I also wanted to isolate it all into one directory in S3 (`/docs`). There’s a kind of dumb `index.html` I’ve manually written that just uses a `<meta>` element to redirect browsers to `/docs`. I *think* I might be able to do this better by setting [`default_root_object`](https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/blob/main/README.md#input_default_root_object) in the Cloudposse module in Terraform, but I’ll experiment with that later.

Part of https://github.com/usdigitalresponse/univaf/issues/1550.

Screenshot:

Screenshot of new documentation site for archives


Mr0grog:
TODO: add info about final Postgres and SQLite dumps

PR #1572, opened 2023-06-16
Shutdown: Replace API with tombstone, 2nd try

Mr0grog:
The previous change (#1571) did not take when Terraform applied it. Hopefully changing the resource name (so we delete the previous record and create a new one) works.

The specific error was:

> Error: updating Route 53 resource record sets: InvalidChangeBatch: [RRSet of type CNAME with DNS name getmyvax.org. is not permitted at apex in zone getmyvax.org.] status code: 400, request id: 14b91d13-0fc8-4e6f-b70c-606f403e8965
> with aws_route53_record.api_domain_record[0]
> on api-domains.tf line 38, in resource "aws_route53_record" "api_domain_record":

So it may also turn out that we can't use a CNAME here in Route 53, but that seems surprising. We’ll see.

Part of #1550.

### DIFF
--- a/terraform/api-domains.tf
+++ b/terraform/api-domains.tf
@@ -35,18 +35,13 @@ data "aws_route53_zone" "domain_zone" {
 }
 
 # DNS record for the domain specified in the `domain_name` variable.
-resource "aws_route53_record" "api_domain_record" {
-  count = var.domain_name != "" ? 1 : 0
-
+resource "aws_route53_record" "api_apex_domain_record" {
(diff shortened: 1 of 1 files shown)


Mr0grog:
Hmmm, Terraform seems to have forgotten about the old record and is not planning a deletion here…

Mr0grog:
Yep, it's just lost from the state — the putting back the old configuration is causing Terraform to plan to create a new record, even though that record exists. So! I am just going to delete the current apex record in the AWS web console, and create the new CNAME via Terraform here.

PR #1577, opened 2023-06-16
Shutdown: Destroy API server

Mr0grog:
The API server is no longer reachable, so it can safely no longer exist. This also stops auto-deploying in our CI workflow and auto-publishing docker images to ECR.

Part of #1550.


Mr0grog:
Fingers crossed that Terraform resolves dependencies correctly and tears things down without causing errors!

Mr0grog:
Amazingly, it worked.

PR #1582, opened 2023-06-19
Shutdown: Delete database and bastion

Mr0grog:
We've archived final dumps of the database in S3 at archives.getmyvax.org, and all the services and resources that touch the database been been removed. It's time to drop the database, too. That leaves us without any reason for having the bastion server, so this also removes it.

Part of https://github.com/usdigitalresponse/univaf/issues/1550.


Mr0grog:
Rebased after the mess with ECR.

PR #1583, opened 2023-06-19
Shutdown: Prep ECR for deletion

Mr0grog:
#1581 did not succeed because the ECR repositories are not empty. I think we need. to set `force_delete` on them, *then* delete them. But first we'll see if restoring the configuration is going to cause Terraform to try and recreate them, or if it just puts things back into a good state.

If it causes Terraform to try and recreate them, I'll close this and just delete them manually.

### DIFF
--- a/terraform/ecr.tf
+++ b/terraform/ecr.tf
@@ -28,6 +28,7 @@ locals {
 resource "aws_ecr_repository" "server_repository" {
   name                 = "univaf-server"
   image_tag_mutability = "IMMUTABLE"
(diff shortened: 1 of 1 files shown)


Mr0grog:
This needs to recreated the lifecycle policies (which *did* get successfully deleted), but correctly knows this means no changes to the repositories themselves.

I've added another commit to use force_delete, which I think will prep the repositories to delete correctly in the next PR.

Mr0grog:
Part of #1550.

PR #1584, opened 2023-06-19
Shutdown: Drop ECR repositories, take 2

Mr0grog:
This is a follow-on to #1581 and #1583. The original deletion of our ECR repos didn't take because they weren't empty and we hadn't configured them with force_delete. #1583 should have configured them so deleting them now works in this PR.

Part of #1550.


Mr0grog:
🤞 :shipit:

Mr0grog:
Success!

PR #1586, opened 2023-06-19
Shutdown: Use S3 bucket name variable

Mr0grog:
Follow-on to #1585. Not sure why we weren't using this variable before. But we're using it now!

Part of #1550.


Mr0grog:
No concrete changes for Terraform to do, which is exactly as expected. Pure refactor.

PR #1592, opened 2023-10-19
Revert "build(deps-dev): bump @babel/traverse from 7.22.4 to 7.23.2"

astonm:
This reverts commit 797ad5223c18a52557ab7c854461f02ad50f4bad.

I'm hoping it repairs the currently-broken build.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1381,13 +1381,12 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.22.13",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.22.13.tgz",
-      "integrity": "sha512-XktuhWlJ5g+3TJXc5upd9Ks1HutSArik6jf2eAjYFyIOf4ej3RN+184cZbzDvbPnuTJIUhPKKJE3cIsYTiAT3w==",
+      "version": "7.21.4",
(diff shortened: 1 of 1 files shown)


Mr0grog:
Ah, looks like the problem isn’t actually 797ad52 — it’s just that [the Dockerfile is based on Node.js 18.16.0](https://github.com/usdigitalresponse/univaf/blob/797ad5223c18a52557ab7c854461f02ad50f4bad/Dockerfile#L1), and then tries to update NPM to the latest version, which requires >= 18.17.0. So this is just down to the passage of time. From the build log:

```
 > [base 2/9] RUN npm install -g npm:
 ERR! code EBADENGINE
0.998 npm ERR! engine Unsupported engine
0.998 npm ERR! engine Not compatible with your version of node/npm: npm@10.2.1
0.999 npm ERR! notsup Not compatible with your version of node/npm: npm@10.2.1
0.999 npm ERR! notsup Required: {"node":"^18.17.0 || >=20.5.0"}
0.999 npm ERR! notsup Actual:   {"npm":"9.5.1","node":"v18.16.0"}
```

The right fix is probably to update the Node.js version in the Dockerfile to the latest 18.x instead.

But also the site is shut down and just running static content, so this hasn’t caused any real problems, so no worries!

Mr0grog:
Should be solved in #1593; I’m closing this.

